PR #5, opened 2018-11-29
Init gradle build

birdstorm:
Initial gradle build for #2


build.gradle
```diff
@@ -0,0 +1,154 @@
...
+			include '**/*.proto'
+        }
+        java {
+            // include self written and generated code
+            srcDirs 'src/main/java', 'target/generated-sources/main/java'            
+        }
+    }
+	test {
+		java {
+			srcDirs 'src/test/java'
```
ice1000: This srcDir is added by default.

build.gradle
```diff
@@ -0,0 +1,154 @@
+apply plugin: 'java'
+apply plugin: 'maven'
+apply plugin: 'idea'
```
ice1000: Is this really used?

birdstorm:
PTAL @ice1000 thx~

birdstorm:
yeah, I have published the original tikv-client to maven, and I think I should do the same for this raw client.

ice1000:
Cool!

PR #14, opened 2018-12-17
Extend for YCSB

birdstorm:


src/main/java/org/tikv/common/exception/KeyException.java
```diff
@@ -26,6 +29,7 @@ public KeyException(String errMsg) {
   }
```
ice1000: Better to change this into `this(null)`.

src/main/java/org/tikv/raw/RawKVClient.java
```diff
@@ -1,20 +1,38 @@
...
 import org.tikv.common.util.BackOffer;
 import org.tikv.common.util.ConcreteBackOffer;
 import org.tikv.common.util.Pair;
 import org.tikv.kvproto.Kvrpcpb;
 import org.tikv.kvproto.Metapb;
 
-public class RawKVClient {
+import java.util.*;
+
+public class RawKVClient implements AutoCloseable {
```
ice1000: Why not implement `Closeable` too?

src/test/java/org/tikv/raw/RawKVClientTest.java
```diff
@@ -152,7 +164,18 @@ private void prepare() {
...
+    remainingKeys.forEach(
+        kvPair -> {
+          try {
+            checkDelete(kvPair.getKey());
+          } catch (final TiKVException e) {
+            logger.warn(
+                "kvPair check delete fails: "
+                    + KeyUtils.formatBytes(kvPair.getKey())
+                    + ", "
+                    + KeyUtils.formatBytes(kvPair.getValue()));
```
ice1000: Can we have a function that wraps this `KeyUtils.formatBytes(key) + ", " + KeyUtils.formatBytes(val)` logic?

src/main/java/org/tikv/raw/RawKVClient.java
ilovesoup: Shouldn't do both. Closeable extends AutoCluseable with more specific exception type (IOException). Closeable is kind of legacy interface for old version of JDK. Most of the exception thrown from inside session is not IOException, more specifically,  we here prefer unchecked exception.

ice1000: Sounds good.

src/main/java/org/tikv/common/exception/KeyException.java
ice1000: > The other ctor explicitly called super already.
This is why we should call the other ctor which has called super, why "so this is not right"?

ilovesoup: this(null) in turns calls
```
public KeyException(Kvrpcpb.KeyError keyErr) {
    super("Key exception occurred");
    this.keyErr = keyErr;
  }
```
which initialize error message to "Key exception occurred".
If really want to optimize this, there should be a ctor like:
```
public KeyException(String errMsg, Kvrpcpb.KeyError keyErr) {
    super(errMsg);
    this.keyErr = keyErr;
}
```
And other ctor in turns calls this one.

ice1000: Woo! I didn't notice that. Thanks for pointing it out

src/main/java/org/tikv/raw/RawKVClient.java
```diff
@@ -36,16 +54,29 @@ public static RawKVClient create(String address) {
...
    * @param value raw value
    */
   public void put(ByteString key, ByteString value) {
-    Pair<TiRegion, Metapb.Store> pair = regionManager.getRegionStorePairByKey(key);
-    RegionStoreClient client = RegionStoreClient.create(pair.first, pair.second, session);
-    client.rawPut(defaultBackOff(), key, value);
+    BackOffer backOffer = defaultBackOff();
+    while (true) {
+      Pair<TiRegion, Metapb.Store> pair = regionManager.getRegionStorePairByKey(key);
+      RegionStoreClient client = RegionStoreClient.create(pair.first, pair.second, session);
```
ilovesoup: For a lightweight action like put, reconstruct client each time might cost too much on itself and GC. Likely we put a todo here and cache client later if we have time.

src/main/java/org/tikv/raw/RawKVClient.java
```diff
@@ -36,16 +54,29 @@ public static RawKVClient create(String address) {
...
    *
    * @param key raw key
    * @param value raw value
    */
   public void put(ByteString key, ByteString value) {
-    Pair<TiRegion, Metapb.Store> pair = regionManager.getRegionStorePairByKey(key);
-    RegionStoreClient client = RegionStoreClient.create(pair.first, pair.second, session);
-    client.rawPut(defaultBackOff(), key, value);
+    BackOffer backOffer = defaultBackOff();
+    while (true) {
```
ilovesoup: Is while (true) too brutal? Shall we have a limit here?

src/main/java/org/tikv/raw/RawKVClient.java
```diff
@@ -54,25 +85,35 @@ public void put(ByteString key, ByteString value) {
...
-    for (Map.Entry<Pair<TiRegion, Metapb.Store>, List<Kvrpcpb.KvPair>> entry :
-        regionMap.entrySet()) {
-      RegionStoreClient client =
-          RegionStoreClient.create(entry.getKey().first, entry.getKey().second, session);
-      if (!client.rawBatchPut(defaultBackOff(), entry.getValue())) {
-        remainingPairs.addAll(entry.getValue());
+      for (Map.Entry<Pair<TiRegion, Metapb.Store>, List<Kvrpcpb.KvPair>> entry :
+          regionMap.entrySet()) {
+        RegionStoreClient client =
+            RegionStoreClient.create(entry.getKey().first, entry.getKey().second, session);
```
ilovesoup: TODO: we need to double check how grpc itself manage connection resource.

src/main/java/org/tikv/raw/RawKVClient.java
```diff
@@ -54,25 +85,35 @@ public void put(ByteString key, ByteString value) {
...
+      if (remainingPairs.isEmpty()) {
+        return;
       }
-    }
-    if (!remainingPairs.isEmpty()) {
       // re-splitting ranges
-      batchPut(remainingPairs);
+      backOffer.doBackOff(
+          BackOffFunction.BackOffFuncType.BoRegionMiss,
+          new TiKVException("BatchPut encounter exception, need re-split the ranges"));
```
ilovesoup: why we need another exception type here?

PR #16, opened 2018-12-19
Add multi-thread for batchPut

birdstorm:


src/main/java/org/tikv/raw/RawKVClient.java
```diff
@@ -30,16 +34,21 @@
...
+  private static final Logger logger = Logger.getLogger(RawKVClient.class);
+
+  private static final int RAW_BATCH_PUT_SIZE = 16 * 1024;
 
   private RawKVClient(String addresses) {
     session = TiSession.create(TiConfiguration.createRawDefault(addresses));
     regionManager = session.getRegionManager();
+    ExecutorService executors =
+        Executors.newFixedThreadPool(session.getConf().getRawClientConcurrency());
+    completionService = new ExecutorCompletionService<>(executors);
```
ilovesoup: Might better consider using newCachedThreadPool. If user don't use it, no need to pay the resource of it.

src/main/java/org/tikv/raw/RawKVClient.java
```diff
@@ -79,42 +88,32 @@ public void put(ByteString key, ByteString value) {
...
-    while (true) {
-      kvPairs.addAll(remainingPairs);
-      remainingPairs.clear();
-      Map<Pair<TiRegion, Metapb.Store>, List<Kvrpcpb.KvPair>> regionMap = new HashMap<>();
-      for (Kvrpcpb.KvPair kvPair : kvPairs) {
-        Pair<TiRegion, Metapb.Store> pair = regionManager.getRegionStorePairByKey(kvPair.getKey());
-        regionMap.computeIfAbsent(pair, t -> new ArrayList<>()).add(kvPair);
-      }
+  public void batchPut(BackOffer backOffer, List<ByteString> keys, List<ByteString> values) {
+    assert keys.size() == values.size();
```
ilovesoup: Use guava's assert. Unless manually turned on, assert keyword will not be active in production. What you actually need here is indeed an exception.
Or consider use something like Pair then you don't need such check / or a Map since a key should not be insert twice.

src/main/java/org/tikv/raw/RawKVClient.java
```diff
@@ -190,6 +189,123 @@ public void delete(ByteString key) {
...
+   */
+  private void sendBatchPut(BackOffer backOffer, List<Batch> batches) {
+    for (Batch batch : batches) {
+      completionService.submit(
+          () -> {
+            Pair<TiRegion, Metapb.Store> pair =
+                regionManager.getRegionStorePairByRegionId(batch.regionId);
+            RegionStoreClient client = RegionStoreClient.create(pair.first, pair.second, session);
+            List<Kvrpcpb.KvPair> kvPairs = new ArrayList<>();
+            for (int i = 0; i < batch.keys.size(); i++) {
```
ilovesoup: Can we directly use List<KvPair> inside of batch? It's kind of meaningless to copy twice.

src/main/java/org/tikv/raw/RawKVClient.java
```diff
@@ -190,6 +189,123 @@ public void delete(ByteString key) {
...
+                      .build());
+            }
+            try {
+              client.rawBatchPut(backOffer, kvPairs);
+            } catch (final TiKVException e) {
+              // TODO: any elegant way to re-split the ranges if fails?
+              backOffer.doBackOff(BackOffFunction.BackOffFuncType.BoRegionMiss, e);
+              logger.warn("ReSplitting ranges for BatchPutRequest");
+              // recursive calls
+              batchPut(backOffer, batch.keys, batch.values);
```
ilovesoup: Looks a little bit strange to call batchPut directly here.

src/main/java/org/tikv/raw/RawKVClient.java
birdstorm: done.

src/main/java/org/tikv/raw/RawKVClient.java
birdstorm: done.

PR #26, opened 2019-01-21
fix gradle build bugs, remove hadoop dependencies

honeychen502:
Signed-off-by: jackycchen <120505816@qq.com>
Fix two things:
1. The major thing, fix gradle build file bugs, and we can use "gradle clean build -x test" run successfully.
2. Secondly, optimize ConcreteBackOffer somethings according to go-client code.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -25,12 +25,12 @@ buildscript {
 sourceSets {
     main {
         proto {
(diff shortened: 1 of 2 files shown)


birdstorm:
Great job! Thanks again for your neat patch! @honeychen502

PR #28, opened 2019-01-28
2PC pull request

honeychen502:
New:
1. 2pc logic implements, according to go-client 2pc. Code location: txn package.
2. add some txn api in RegionStoreClient file.
3. some unit test cases for 2pc.

Not:
There are some conflicts that haven't be solved. For example, Snapshot class is still used in my branch.
Problem is :  Is there still necessary to keep Snapshot file for transaction snapshot reading.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -25,12 +25,12 @@ buildscript {
 sourceSets {
     main {
         proto {
-            srcDir './proto'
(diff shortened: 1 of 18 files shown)


ice1000:
Why is this closed?

PR #48, opened 2019-02-19
Add GC worker

birdstorm:
A simple implementation of GC worker for java client

### DIFF
--- a/src/main/java/org/tikv/common/PDClient.java
+++ b/src/main/java/org/tikv/common/PDClient.java
@@ -30,6 +30,7 @@ import java.util.concurrent.Future;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.TimeUnit;
 import java.util.function.Supplier;
+import java.util.stream.Collectors;
 import org.tikv.common.TiConfiguration.KVMode;
 import org.tikv.common.codec.Codec.BytesCodec;
 import org.tikv.common.codec.CodecDataOutput;
@@ -48,17 +49,6 @@ import org.tikv.kvproto.PDGrpc;
 import org.tikv.kvproto.PDGrpc.PDBlockingStub;
 import org.tikv.kvproto.PDGrpc.PDStub;
 import org.tikv.kvproto.Pdpb.*;
-
-import java.net.MalformedURLException;
-import java.net.URI;
-import java.net.URL;
-import java.util.List;
-import java.util.concurrent.Executors;
-import java.util.concurrent.Future;
-import java.util.concurrent.ScheduledExecutorService;
-import java.util.concurrent.TimeUnit;
-import java.util.function.Supplier;
-import java.util.stream.Collectors;
 
 import static com.google.common.base.Preconditions.checkNotNull;
 import static org.tikv.common.operation.PDErrorHandler.getRegionResponseErrorExtractor;
--- a/src/main/java/org/tikv/common/PDClient.java
+++ b/src/main/java/org/tikv/common/PDClient.java
@@ -15,28 +15,19 @@
 
 package org.tikv.common;
 
-import static com.google.common.base.Preconditions.checkNotNull;
-import static org.tikv.common.operation.PDErrorHandler.getRegionResponseErrorExtractor;
-import static org.tikv.common.pd.PDError.buildFromPdpbError;
-
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.net.HostAndPort;
 import com.google.common.util.concurrent.ThreadFactoryBuilder;
 import com.google.protobuf.ByteString;
 import io.etcd.jetcd.*;
 import io.etcd.jetcd.kv.GetResponse;
+import io.etcd.jetcd.lease.LeaseGrantResponse;
 import io.etcd.jetcd.lock.LockResponse;
-import io.etcd.jetcd.lock.UnlockResponse;
 import io.grpc.ManagedChannel;
-import java.net.MalformedURLException;
-import java.net.URL;
-import java.util.List;
-import java.util.concurrent.*;
-import java.util.function.Supplier;
-
 import org.tikv.common.TiConfiguration.KVMode;
 import org.tikv.common.codec.Codec.BytesCodec;
 import org.tikv.common.codec.CodecDataOutput;
+import org.tikv.common.exception.GCException;
 import org.tikv.common.exception.GrpcException;
 import org.tikv.common.exception.TiClientInternalException;
 import org.tikv.common.exception.TiKVException;
@@ -52,7 +43,24 @@ import org.tikv.kvproto.PDGrpc.PDBlockingStub;
 import org.tikv.kvproto.PDGrpc.PDStub;
 import org.tikv.kvproto.Pdpb.*;
 
-/** PDClient is thread-safe and suggested to be shared threads */
+import java.net.MalformedURLException;
+import java.net.URI;
+import java.net.URL;
+import java.util.List;
+import java.util.concurrent.Executors;
+import java.util.concurrent.Future;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.TimeUnit;
+import java.util.function.Supplier;
+import java.util.stream.Collectors;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+import static org.tikv.common.operation.PDErrorHandler.getRegionResponseErrorExtractor;
+import static org.tikv.common.pd.PDError.buildFromPdpbError;
+
+/**
+ * PDClient is thread-safe and suggested to be shared threads
+ */
 public class PDClient extends AbstractGRPCClient<PDBlockingStub, PDStub>
     implements ReadOnlyPDClient {
   private RequestHeader header;
@@ -165,10 +173,13 @@ public class PDClient extends AbstractGRPCClient<PDBlockingStub, PDStub>
     return responseObserver.getFuture();
   }
 
+  private Supplier<GetStoreRequest> getStoreRequest(long storeId) {
+    return () -> GetStoreRequest.newBuilder().setHeader(header).setStoreId(storeId).build();
+  }
+
   @Override
   public Store getStore(BackOffer backOffer, long storeId) {
-    Supplier<GetStoreRequest> request =
-        () -> GetStoreRequest.newBuilder().setHeader(header).setStoreId(storeId).build();
+    Supplier<GetStoreRequest> request = getStoreRequest(storeId);
     PDErrorHandler<GetStoreResponse> handler =
         new PDErrorHandler<>(
             r -> r.getHeader().hasError() ? buildFromPdpbError(r.getHeader().getError()) : null,
@@ -183,8 +194,7 @@ public class PDClient extends AbstractGRPCClient<PDBlockingStub, PDStub>
     FutureObserver<Store, GetStoreResponse> responseObserver =
         new FutureObserver<>(GetStoreResponse::getStore);
 
-    Supplier<GetStoreRequest> request =
-        () -> GetStoreRequest.newBuilder().setHeader(header).setStoreId(storeId).build();
+    Supplier<GetStoreRequest> request = getStoreRequest(storeId);
     PDErrorHandler<GetStoreResponse> handler =
         new PDErrorHandler<>(
             r -> r.getHeader().hasError() ? buildFromPdpbError(r.getHeader().getError()) : null,
@@ -218,8 +228,8 @@ public class PDClient extends AbstractGRPCClient<PDBlockingStub, PDStub>
     return leaderWrapper;
   }
 
-  private String getLeaderUrl() {
-    return "http://" + getLeaderWrapper().getLeaderInfo();
+  private List<URI> getMemberUrI() {
+    return pdAddrs.stream().map(x -> URI.create("http://" + x.toString())).collect(Collectors.toList());
   }
 
   public ByteString get(ByteString key) {
@@ -232,7 +242,7 @@ public class PDClient extends AbstractGRPCClient<PDBlockingStub, PDStub>
       ByteSequence value = kvs.get(0).getValue();
       return ByteString.copyFrom(value.getBytes());
     } catch (Exception e) {
-      throw new TiKVException("Unhandled");
+      throw new TiKVException("get etcd key fails", e);
     }
   }
 
@@ -248,23 +258,24 @@ public class PDClient extends AbstractGRPCClient<PDBlockingStub, PDStub>
     try {
       etcdKV.put(ByteSequence.from(key), ByteSequence.from(value)).get(500, TimeUnit.MILLISECONDS);
     } catch (Exception e) {
-      throw new TiKVException("Unhandled");
+      throw new TiKVException("put etcd key fails", e);
     }
+  }
+
+  public Txn txn() {
+    return etcdKV.txn();
   }
 
   /**
    * Tries to acquire lock with given name from PD
    *
-   * @param key name of lock
+   * @param key   name of lock
    * @param lease lease when lock expires
    * @return whether the lock is acquired successfully
    */
   public boolean lock(ByteString key, long lease) {
     try {
-      logger.info("locking " + key.toStringUtf8() + " lease=" + lease);
       LockResponse resp = lockClient.lock(ByteSequence.from(key), lease).get(500, TimeUnit.MILLISECONDS);
-      logger.info("locked " + key.toStringUtf8());
-      logger.info("returnKey" + ByteString.copyFrom(resp.getKey().getBytes()).toStringUtf8());
       return true;
     } catch (Exception e) {
       logger.warn("lock fails", e);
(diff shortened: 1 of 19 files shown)


birdstorm:
@disksing @zhangjinpeng1987 @AndreMouche Anyone will do, PTAL thx~

src/main/java/org/tikv/common/PDClient.java
```diff
@@ -212,6 +228,115 @@ LeaderWrapper getLeaderWrapper() {
     return leaderWrapper;
   }
 
+  private List<URI> getMemberUrI() {
```
kennytm: This is an odd spelling even for a private API 😅 

```suggestion
  private List<URI> getMemberURI() {
```

birdstorm: thanks for pointing out

disksing:
@MyonKeminta Please take a look

src/main/java/org/tikv/txn/TxnKVClient.java
```diff
@@ -1,5 +1,27 @@
...
 package org.tikv.txn;
 
+import org.tikv.common.TiSession;
+import org.tikv.txn.gc.GCWorker;
+
 public class TxnKVClient {
   // TODO: To be done.
+  public TxnKVClient(TiSession session) {
+    GCWorker worker = new GCWorker(session);
+    worker.run();
```
MyonKeminta: Is the client designed to keep running for a long time?

src/main/java/org/tikv/txn/gc/GCWorker.java
```diff
@@ -0,0 +1,467 @@
...
+  private ExecutorService gcTaskThreadPool;
+  private ExecutorCompletionService<GCTask> gcTaskService;
+
+  private static final ByteString GC_LAST_RUN_TIME_KEY = ByteString.copyFromUtf8("tikv_gc_last_run_time");
+  private static final ByteString GC_RUN_INTERVAL_KEY = ByteString.copyFromUtf8("tikv_gc_run_interval");
+  private static final ByteString GC_LIFE_TIME_KEY = ByteString.copyFromUtf8("tikv_gc_life_time");
+  private static final ByteString GC_SAFE_POINT_KEY = ByteString.copyFromUtf8("tikv_gc_safe_point");
+  private static final ByteString GC_CONCURRENCY_KEY = ByteString.copyFromUtf8("tikv_gc_concurrency");
+  private static final ByteString GC_LEADER_UUID_KEY = ByteString.copyFromUtf8("tikv_gc_leader_uuid");
+  private static final ByteString GC_LEADER_LEASE_KEY = ByteString.copyFromUtf8("tikv_gc_leader_lease");
```
MyonKeminta: Are these keys saved directly in PD's etcdkv? If the client need to work while TiDB is also running in the cluster, I think TiDB's code also need to be changed accordingly.

birdstorm: Yeah I'm also concerned about this. I don't think any TiKV client necessarily need the existence of TiDB. Furthermore, I doubt it can and should work alongside with TiDB because problem occurs when both clients start the GC job.

src/main/java/org/tikv/txn/TxnKVClient.java
birdstorm: No it is not. Current GCWorker in TiDB is not designed for a single client (which is why I have to use a lock for multi-client scenario). In my opinion GCWorker should belong to an independent role rather than be moulded into computing layer. It should keep alive along with TiKV even if the client is not working. PD, for instance, should be a good role for starting GC requests but I understand it cannot by design.

MyonKeminta: Yes! I always want to make GC an internal component of TiKV (and PD) so that users of pure KV won't need to write a client to GC the cluster anymore.
But here... I'm afraid this is not a very good idea if the programs that uses the TxnKVClient might exit soon...

birdstorm: Yeah now that there isn't a good role to implement gc logic for client . Do you have any good ideas so that we can discuss its possibility? 

Here are a few potential choices: 

- start a GCWorker in background and make it it independent to our client (we probably call it GCClient)
- change our client into the same server-client structure as TiDB, and embed GC logic in the server with a much longer lifetime

The problem actually lies here:

GC is only valid when TiKV is in TXN mode. The GCWorker should be notified that it does not work with raw KV operations. Things get a bit difficult when trying to separate their logic from each other when they are concerned to be different components.

MyonKeminta: I believe GC should be an internal functionality of TiKV, but it may need much time to discuss and implement it. Before that I think it's ok to adopt your first choice (which is also my idea).
As for raw kv.. It used to confuse some users that when they put some raw keys in TiKV of a TiDB cluster, TiDB got into failures soon. I think raw kv and txn kv shouldn't be used at the same time. So for clusters uses in raw kv mode,  we simply don't GC it.

birdstorm: Fine, I will change this pr into an independent gc client before we have a better solution.

birdstorm:
@MyonKeminta Hey, I've changed the gc client into an independent client as we have discussed.

src/main/java/org/tikv/txn/gc/GCWorker.java
```diff
@@ -0,0 +1,467 @@
...
+  }
+
+  private void runGCJob(long safePoint) {
+    try {
+      resolveLocks(safePoint);
+    } catch (Exception e) {
+      logger.error(String.format("[gc worker] %s resolve locks returns an error %s", uuid, e));
+      return;
+    }
+    doGC(safePoint);
```
MyonKeminta: From TiKV 3.0, "Distributed GC" has been supported. So for TiKV 3.0 clusters, after resolving locks, you can simply upload the safepoint to PD via PD's "uploadGCSafePoint" API

src/main/java/org/tikv/txn/gc/GCWorker.java
```diff
@@ -0,0 +1,467 @@
...
+    }
+    if (safePoint == 0) {
+      gcIsRunning = false;
+      return;
+    }
+
+    gcIsRunning = true;
+    logger.info(String.format("[gc worker] %s starts the whole job, safePoint: %d", uuid, safePoint));
+    runGCJob(safePoint);
+    lastFinish = System.currentTimeMillis();
```
MyonKeminta: You used system time here but used TSO at L140. How about use the same?

src/main/java/org/tikv/txn/gc/GCWorker.java
```diff
@@ -0,0 +1,467 @@
...
+  }
+
+  private boolean doCheckLeader() {
+    long leader = getLong(GC_LEADER_UUID_KEY);
+    logger.debug(String.format("[gc worker] got leader: %s", Long.toHexString(leader)));
+    if (leader == l_uuid) {
+      try {
+        putLong(GC_LEADER_LEASE_KEY, System.currentTimeMillis() + gcWorkerLease);
+      } catch (Exception e) {
+        return false;
```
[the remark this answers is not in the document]
birdstorm: done

src/main/java/org/tikv/txn/gc/GCWorker.java
birdstorm: done

src/main/java/org/tikv/txn/gc/GCWorker.java
```diff
@@ -0,0 +1,467 @@
...
+    scheduler.shutdown();
+    gcWorkerScheduler.shutdown();
+  }
+
+  private TiTimestamp getTimestamp() {
+    return pdClient.getTimestamp(ConcreteBackOffer.newTsoBackOff());
+  }
+
+  // Note: Should not call run() more than once
+  public void run() {
```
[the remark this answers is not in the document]
birdstorm: done

src/main/java/org/tikv/txn/gc/GCWorker.java
birdstorm: I can make it configurable in another PR.

src/main/java/org/tikv/txn/gc/GCWorker.java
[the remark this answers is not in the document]
birdstorm: donedone

birdstorm:
@MyonKeminta PTAL thx.

src/main/java/org/tikv/txn/gc/GCWorker.java
```diff
@@ -0,0 +1,472 @@
...
+    if (safePoint == 0) {
+      gcIsRunning = false;
+      return;
+    }
+
+    gcIsRunning = true;
+    logger.info(String.format("[gc worker] %s starts the whole job, safePoint: %d", uuid, safePoint));
+    runGCJob(safePoint);
+    lastFinish = getNow();
+    gcIsRunning = false;
```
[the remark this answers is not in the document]
birdstorm: It is async, and I believe we do not need to care about the `gcIsRunning` flag being unset because it only notifies `GCWorker` itself whether a job is running.

src/main/java/org/tikv/txn/gc/GCWorker.java
```diff
@@ -0,0 +1,472 @@
...
+        TiRegion region = pdClient.getRegionByKey(backOffer, key);
+        try {
+          doGCForRegion(backOffer, safePoint, region);
+        } catch (RegionException e) {
+          backOffer.doBackOff(BackOffFunction.BackOffFuncType.BoRegionMiss, e);
+        } catch (Exception e) {
+          logger.warn(String.format("[gc worker] %s gc for range [%s, %s) safepoint: %d, failed", identifier, KeyUtils.formatBytes(startKey), KeyUtils.formatBytes(endKey), safePoint), e);
+        }
+        key = region.getEndKey();
+        if (key.equals(ByteString.EMPTY) || FastByteComparisons.compareTo(key.toByteArray(), endKey.toByteArray()) >= 0) {
```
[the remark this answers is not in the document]
birdstorm: nice catch!

src/main/java/org/tikv/txn/gc/GCWorker.java
```diff
@@ -0,0 +1,472 @@
...
+      try {
+        logger.info("[gc worker] run gc task worker");
+        while (true) {
+          Future<GCTask> gcTaskFuture;
+          if ((gcTaskFuture = gcTaskService.take()) == null) {
+            logger.error("[gc worker] No tasks remain");
+            return;
+          }
+          task = gcTaskFuture.get();
+          if (logger.isDebugEnabled()) {
```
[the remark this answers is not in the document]
birdstorm: to prevent unnecessary formatting when using other logger levels.

src/main/java/org/tikv/txn/gc/GCWorker.java
```diff
@@ -0,0 +1,472 @@
...
+    }
+
+    void run() {
+      GCTask task;
+      try {
+        logger.info("[gc worker] run gc task worker");
+        while (true) {
+          Future<GCTask> gcTaskFuture;
+          if ((gcTaskFuture = gcTaskService.take()) == null) {
+            logger.error("[gc worker] No tasks remain");
```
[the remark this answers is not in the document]
birdstorm: done

src/main/java/org/tikv/txn/gc/GCWorker.java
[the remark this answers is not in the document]
MyonKeminta: If `gcIsRunning` is false, but the GCWorker is actually asynchronously processing GC tasks for each region, won't it went wrong?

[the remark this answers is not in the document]
birdstorm: fixed

birdstorm:
@MyonKeminta Done. PTAL.
I will use the new pd api in next pr.

src/main/java/org/tikv/txn/gc/GCWorker.java
```diff
@@ -0,0 +1,492 @@
...
+  private static final ByteString GC_LEADER_UUID_KEY = ByteString.copyFromUtf8("tikv_gc_leader_uuid");
+  private static final ByteString GC_LEADER_LEASE_KEY = ByteString.copyFromUtf8("tikv_gc_leader_lease");
+  public static final int gcScanLockLimit = 1024;
+  private final int gcDefaultRunInterval;// = 6 * 1000
+  private final int gcDefaultLifeTime;// = 2 * 1000;
+  private final int gcDefaultConcurrency;
+  private static final int gcMinConcurrency = 1;
+  private static final int gcMaxConcurrency = 128;
+  private final int gcWaitTime;// = 3 * 1000;
+  private final int gcWorkerLease;// = 2 * 1000;
```
MyonKeminta: It seems these comments is not consistent with the default value, which is 2 * 60 * 1000 milliseconds?

src/main/java/org/tikv/txn/gc/GCWorker.java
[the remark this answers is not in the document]
MyonKeminta: Did you forgot to update this? 🤣

src/main/java/org/tikv/txn/gc/GCWorker.java
```diff
@@ -0,0 +1,492 @@
...
+      }
+    } catch (Exception e) {
+      awaitGCComplete();
+    }
+  }
+
+  private void awaitGCComplete() {
+    try {
+      if (gcTaskThreadPool != null) {
+        gcTaskThreadPool.awaitTermination(gcWorkerLease, TimeUnit.MILLISECONDS);
```
MyonKeminta: Is the parameter the timeout? If so, this doesn't seems to be correct. It might take long time to finish GC, while the tick should continue. The leader lease should still be updated in leaderTick.

src/main/java/org/tikv/txn/gc/GCWorker.java
```diff
@@ -0,0 +1,492 @@
...
+    logger.info(String.format("[gc worker] %s starts", uuid));
+    scheduler.scheduleAtFixedRate(() -> {
+      try {
+        logger.debug("[gc worker] tick");
+        pdClient.keepLeaseAlive(l_uuid);
+        tick();
+      } catch (Exception e) {
+        logger.warn("[gc worker] gc fails to proceed", e);
+      }
+    }, 0, 1, TimeUnit.SECONDS);
```
MyonKeminta: Here, if a tick doesn't finish for a long time, will the next tick run asynchronously? If it's not so, then the `runGCJob` at L164 should be asynchronous. Sorry I didn't notice this question in my previous review.

src/main/java/org/tikv/txn/gc/GCWorker.java
birdstorm: yeah, i will delete them

src/main/java/org/tikv/txn/gc/GCWorker.java
birdstorm: JavaDoc
> If any execution of this task takes longer than its period, then subsequent executions may start late, but will not concurrently execute.

MyonKeminta: Oh, then the `runGCJob` at L164 should be a async task. When the db is very large it's possible to take very long time to finish the whole gc, and the tick shouldn't stop.

src/main/java/org/tikv/txn/gc/GCWorker.java
birdstorm: Found out that I forgot to make runGCJob async. My bad.

birdstorm: done.

birdstorm:
@MyonKeminta PTAL.

src/main/java/org/tikv/txn/gc/GCWorker.java
MyonKeminta: Emmm, so is the parameters the timeout? It doesn't seems to be correct since it may take very long time to finish.

src/main/java/org/tikv/txn/gc/GCWorker.java
```diff
@@ -0,0 +1,472 @@
...
+
+    private void doGCForRegion(BackOffer backOffer, long safePoint, TiRegion region) {
+      RegionStoreClient client = regionStoreClientBuilder.build(region);
+      logger.debug("[gc worker] doGCForRegion");
+      client.doGC(backOffer, safePoint);
+      logger.debug("[gc worker] ~doGCForRegion");
+    }
+  }
+
+  private GCTaskWorker newGCTaskWorker(String uuid) {
```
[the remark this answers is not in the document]
birdstorm: It creates a new TaskWorker for multithread.

src/main/java/org/tikv/txn/gc/GCWorker.java
birdstorm: There is no other way to identify whether GC is completed or the task executor was lost.

MyonKeminta: Maybe you can add a callback to each GCTask which is invoked on finish, and wait for it to be called enough times?

PR #60, opened 2019-04-10
fix bazel build

zhexuany:
Fix bazel build in this repo.


src/main/java/org/tikv/BUILD
```diff
@@ -1,12 +1,15 @@
...
 java_library(
-    name = "tikv-java-client-lib",
+    name = "tikv_java_client_lib",
     srcs = glob(
         ["**/*.java"],
     ),
     deps = [
-        "//:java",
+        "//:tikv_proto_java_lib",
+#        ":tikv_parser_lib",
```
zhexuany: comment for now since soon or later we will merge parser code into this repo.

birdstorm:
Can you please add some documents for `bazel`? I observe the usage of `bazelisk` rather than `bazel` in your code. @zhexuany

zhexuany:
@birdstorm PTAL

birdstorm:
Closed since we are not going to support bazel.

PR #62, opened 2019-04-28
expose columnfamily as parameter

jaguarx:
Close #62 
Expose column family as parameter in raw put/get/scan


birdstorm:
Hi, @jaguarx ! Thanks for your help.
You should use `git commit --signoff` to pass the DCO check

PR #63, opened 2019-06-14
Fix ScanIterator may fail when an error encounters

birdstorm:
When ScanIterator gets an Error from TiKV indicating region info needs to be refreshed, e.g., region epoch is not match, we should retry with a new client to prevent unrecoverable failure.

### DIFF
--- a/src/main/java/org/tikv/common/operation/iterator/RawScanIterator.java
+++ b/src/main/java/org/tikv/common/operation/iterator/RawScanIterator.java
@@ -38,23 +38,22 @@ public class RawScanIterator extends ScanIterator {
   }
 
   TiRegion loadCurrentRegionToCache() throws Exception {
-    TiRegion region;
-    try (RegionStoreClient client = builder.build(startKey)) {
-      region = client.getRegion();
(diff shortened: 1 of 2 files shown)


src/main/java/org/tikv/common/operation/iterator/RawScanIterator.java
```diff
@@ -38,23 +38,22 @@ public RawScanIterator(
...
-    try (RegionStoreClient client = builder.build(startKey)) {
-      region = client.getRegion();
-      BackOffer backOffer = ConcreteBackOffer.newScannerNextMaxBackOff();
-      if (limit <= 0) {
-        currentCache = null;
-      } else {
-        while (true) {
+    BackOffer backOffer = ConcreteBackOffer.newScannerNextMaxBackOff();
+    while (true) {
+      try (RegionStoreClient client = builder.build(startKey)) {
```
ilovesoup: Is this to fix during retry we need to re-get client since cache might change?

zhexuany: Yes. As @birdstorm stated in the description.

PR #64, opened 2019-08-08
only include proto files, and raft-rs moved proto folder 1 deeper

camerondavison:
fixes protoc java generation


marsishandsome:
@camerondavison Can you use maven to build the project? Cause we are planning to remove Gradle building support.

camerondavison:
yes it does look like the `mvn clean install` command worked for me

PR #72, opened 2019-11-28
add rawBatchGet method

mysunnyshine:
add RawBatchGet method


src/main/java/org/tikv/common/region/RegionStoreClient.java
```diff
@@ -379,6 +381,67 @@ private void handleRawBatchPut(RawBatchPutResponse resp) {
...
+    return rawBatchGetHelper(resp, backOffer);
+  }
+
+  private List<KvPair> rawBatchGetHelper(RawBatchGetResponse resp, BackOffer bo) {
+    List<Lock> locks = new ArrayList<>();
+
+    for (KvPair pair : resp.getPairsList()) {
+      if (pair.hasError()) {
+        if (pair.getError().hasLocked()) {
+          Lock lock = new Lock(pair.getError().getLocked());
```
birdstorm: I think raw batch get will not see lock?

PR #77, opened 2019-12-30
Fix gradle build with shading

birdstorm:


zhexuany:
How can we make sure the correctness of shadeJar generated by Gradle?

PR #87, opened 2020-09-11
fix qps is 0 for 10 minutes when node is down

### LINKED ISSUE #86: Tikv crashes a data node, client writes are unavailable.
Hi, i use java client write and read tikv cluster. I find when one tikv data node crash,client write all data nodes are all unavailable. I need to restart applicaion.

I think when on data node crash, in theory, it should only affect the write of the crash node.

@birdstorm

birdstorm:
Fix #86


marsishandsome (approved):
LGTM

PR #90, opened 2020-09-11
RawGet changes the while (true) mechanism to a retry limit.

### LINKED ISSUE #86: Tikv crashes a data node, client writes are unavailable.

mikechengwei:
Fix #86 

RawGet changes the while (true) mechanism to a retry limit.

### Check List 

Tests 

 - Manual test 
    - Batch execution of RawGet verification.


birdstorm (approved):
LGTM

PR #94, opened 2020-10-23
Remove bazel support from document

birdstorm:
Remove bazel support from document.


marsishandsome (approved):
LGTM

PR #97, opened 2020-10-30
Fix incorrect date type value when timezone is set

birdstorm:
When timezone is set to positive, the date type value may be incorrectly set with `+1` day


marsishandsome (approved):
LGTM

PR #106, opened 2021-01-08
Fix enum index out of range error

birdstorm:
When enum inserted is empty string `''`, its index value is actually 0.

https://dev.mysql.com/doc/refman/8.0/en/enum.html#enum-indexes


marsishandsome (approved):
LGTM

PR #108, opened 2021-01-11
Fix zero in date parsing

birdstorm:
When zero month/day encounters in date, parsing will throw an exception.

Fixes by using DateFormat to parse beforehand.


marsishandsome (approved):
LGTM

PR #110, opened 2021-01-11
Fix empty enum type

birdstorm:
#106  should be bounded -1 rather than 0, because the value of idx was substracted before.


marsishandsome (approved):
LGTM

PR #111, opened 2021-01-11
Fix date time type decoding when contains zero

birdstorm:
per #108 

The same logic shall be also ported to `Codec.decode` apis.


marsishandsome (approved):
LGTM

PR #112, opened 2021-01-14
Fix falsely shaded dependencies

birdstorm:
The `maven-assembly` plugin falsely re-introduces shaded dependencies.

It is now removed in this PR.


marsishandsome (approved):
LGTM

PR #113, opened 2021-01-14
Add raw scan api without limit

birdstorm:
Current APIs for `rawScan` limits the maximum keys to be scanned.

Add a new API to remove this limit. 


marsishandsome (approved):
LGTM

PR #114, opened 2021-01-15
Fix getRegionByKey when using raw mode

birdstorm:
In raw mode, the keys should not be encoded when calling `getRegionByKey()`.


marsishandsome (approved):
LGTM

PR #115, opened 2021-01-18
Add raw scan test with empty result

birdstorm:
Add unit test for `rawScan` with empty result.


marsishandsome (approved):
LGTM

PR #116, opened 2021-01-20
Modify TiSession API document

birdstorm:
As shown in examples, users should use `TiSession.getInstance` rather than `TiSession.create` to initialize a `TiSession`. 


marsishandsome (approved):
LGTM

PR #117, opened 2021-01-27
do not use TiSession.getInstance

marsishandsome:


sunxiaoguang (commented):
Maybe we can eliminate TiSession::getInstance so that people will never have chance to use it wrong.

marsishandsome:
> Maybe we can eliminate TiSession::getInstance so that people will never have chance to use it wrong.

`TiSession::getInstance` is delete from document. I prefer to keep it for the case of compatibility.

sunxiaoguang:
> > Maybe we can eliminate TiSession::getInstance so that people will never have chance to use it wrong.
> 
> `TiSession::getInstance` is delete from document. I prefer to keep it for the case of compatibility.

Sure, let's annotate it as deprecated so we can eventually remove it in the future.

birdstorm (approved):
LGTM

sunxiaoguang:
LGTM

sunxiaoguang (approved):

PR #124, opened 2021-02-04
Retry when error is raft proposal dropped

birdstorm:
If the error message is `Raft Proposal Dropped`, it means that the store is in the middle of a transition. It should be able resolved by a retry.

### DIFF
--- a/src/main/java/org/tikv/common/operation/KVErrorHandler.java
+++ b/src/main/java/org/tikv/common/operation/KVErrorHandler.java
@@ -229,10 +229,14 @@ public class KVErrorHandler<RespT> implements ErrorHandler<RespT> {
         throw new StatusRuntimeException(Status.UNKNOWN.withDescription(error.toString()));
       }
 
-      logger.warn(String.format("Unknown error %s for region [%s]", error.toString(), ctxRegion));
+      logger.warn(String.format("Unknown error %s for region [%s]", error, ctxRegion));
       // For other errors, we only drop cache here.
       // Upper level may split this task.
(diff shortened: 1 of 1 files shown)


src/main/java/org/tikv/common/operation/KVErrorHandler.java
```diff
@@ -229,10 +229,14 @@ public boolean handleResponseError(BackOffer backOffer, RespT resp) {
...
       }
 
-      logger.warn(String.format("Unknown error %s for region [%s]", error.toString(), ctxRegion));
+      logger.warn(String.format("Unknown error %s for region [%s]", error, ctxRegion));
       // For other errors, we only drop cache here.
       // Upper level may split this task.
       invalidateRegionStoreCache(ctxRegion);
+      // retry if raft proposal is dropped, it indicates the store is in the middle of transition
+      if (error.getMessage().contains("Raft Proposal Dropped")) {
+        return true;
```
marsishandsome, line 238: `invalidateRegionStoreCache` should be called in this case?

birdstorm: it has been called above.

marsishandsome: i mean `invalidateRegionStoreCache` should not be called in this case.

birdstorm: it should be called, because the region info is incorrect.

marsishandsome (approved):
LGTM

PR #125, opened 2021-02-18
add random to function getNextFollower

marsishandsome:
update function `getNextFollower`
- add random
- `public` -> `private`
- return `Peer` -> `void`


birdstorm (approved):
LGTM

PR #127, opened 2021-02-20
Modify default RawKV backoff time

birdstorm:
Corresponds with the same configurations in Client-Go.


marsishandsome (approved):
LGTM

PR #131, opened 2021-02-22
Fix cache missing because of recreation of RegionManager

birdstorm:


marsishandsome (approved):
LGTM

PR #133, opened 2021-02-25
Fix retry when TiKV halts

birdstorm:
When TiKV was killed and halts, the default timeout of gRPC message is 10 minutes. So it will wait for 10 minutes to recover to normal QPS.

Fixes by changing the default timeout to 20 seconds.

### DIFF
--- a/src/main/java/org/tikv/common/PDClient.java
+++ b/src/main/java/org/tikv/common/PDClient.java
@@ -500,7 +500,7 @@ public class PDClient extends AbstractGRPCClient<PDBlockingStub, PDStub>
     }
     return leaderWrapper
         .getBlockingStub()
-        .withDeadlineAfter(getConf().getTimeout(), getConf().getTimeoutUnit());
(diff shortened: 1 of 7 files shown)


src/main/java/org/tikv/common/PDClient.java
```diff
@@ -500,17 +500,15 @@ protected PDBlockingStub getBlockingStub() {
     }
     return leaderWrapper
         .getBlockingStub()
-        .withDeadlineAfter(getConf().getTimeout(), getConf().getTimeoutUnit());
+        .withDeadlineAfter(getTimeout(), getConf().getTimeoutUnit());
```
marsishandsome, line 503: `getTimeout(), getConf().getTimeoutUnit()` is a little strange.

both `getConf().getTimeout(), getConf().getTimeoutUnit()` and `getTimeout(), getTimeoutUnit()` are ok.

marsishandsome (commented):
others LGTM

PR #134, opened 2021-02-25
Fix batch put StackOverflow

birdstorm:
batch put may cause StackOverflow because of retry strategy.


marsishandsome (approved):
LGTM

PR #136, opened 2021-02-27
Revert small lock optimization

birdstorm:
small lock optimization does not have a significant effect, thus we revert it for now.


marsishandsome (approved):
LGTM

PR #137, opened 2021-02-27
Optimize group region by keys logic

birdstorm:
In the previous implementation, each key to be grouped should go through the getRegionByKey logic.

But since we know the regions changed which locates the keys in advance, those keys in the same region do not need to go through the same logic again. 


birdstorm:
merged.

PR #140, opened 2021-03-11
Add atomic api support

birdstorm:
Add atomic API support for raw client

Including:
- putIfAbsent
- batchPut
- batchDelete


Little-Wallace (approved):
LGTM

PR #143, opened 2021-03-17
Add Network Mapping

birdstorm:
Add network mapping for Java Client.

By using `-Dtikv.network.mapping=<config-name>`, we can make the client use the corresponding host mapping strategy stored in PD's ETCD in advance.

The format of KV is as follows:
```
key: /client/url-mapping/<conf-name>/<Inet-ip>
value: <mapped-ip>
```


Little-Wallace (approved):
LGTM

PR #145, opened 2021-03-22
Fix batch retry hanging when tikv is down

birdstorm:
When all TiKVs are down, the batch process may hang. And can not recover before the TiKV is up. The backoff did not work properly in this situation.

Fixes this behavior by throwing an exception when the requests wait till backoff time.


Little-Wallace (approved):
LGTM

PR #151, opened 2021-03-26
add replica selector

marsishandsome:


src/main/java/org/tikv/common/InternalReplicaSelector.java
```diff
@@ -0,0 +1,48 @@
...
+
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+import org.tikv.kvproto.Metapb;
+
+public class InternalReplicaSelector implements ReplicaSelector {
+  private final TiConfiguration.ReplicaRead replicaRead;
+
+  public InternalReplicaSelector(TiConfiguration.ReplicaRead replicaRead) {
```
birdstorm, line 26: How about we just give the users 3 internal selectors instances?

e.g., InternalReplicaSelector.LEADER_ONLY_SELECTOR

marsishandsome: fixed

birdstorm (approved):
LGTM

Issue #154, opened 2021-03-28
Building for JDK9+


plevart:
I noticed that JDK 8 is still needed for building. Is this intentional? Do you plan to support JDK 8 in the java-client? I'm looking at some warnings like:
```
[WARNING] /home/peter/work/git/tikv-client-java/src/main/java/org/tikv/common/util/MemoryUtil.java:[31,16] sun.misc.Cleaner is internal proprietary API and may be removed in a future release
[WARNING] /home/peter/work/git/tikv-client-java/src/main/java/org/tikv/common/util/MemoryUtil.java:[32,16] sun.misc.Unsafe is internal proprietary API and may be removed in a future release
[WARNING] /home/peter/work/git/tikv-client-java/src/main/java/org/tikv/common/util/MemoryUtil.java:[33,18] sun.nio.ch.DirectBuffer is internal proprietary API and may be removed in a future release
[WARNING] /home/peter/work/git/tikv-client-java/src/main/java/org/tikv/common/util/FastByteComparisons.java:[23,16] sun.misc.Unsafe is internal proprietary API and may be removed in a future release
```
There are alternatives in JDK 9+ that don't produce such warnings which are now runtime errors in JDK 16+... How much do you value being able to run on JDK 8 ? There is an option of creating alternative classes such as MemoryUtil for JDK 8 vs. JDK 9+ and pack them into a multi-release JAR which would run correctly across all the JDK 8, 9, ..., 16+ releases. I can try to do that and propose a PR if you're interested.
The prerequisite for doing that is to require JDK 9+ (preferably JDK 11) to build the project via maven. The javac would use -release 8 option then to build most of the code so it would run on JDK 8, only those classes that need to use different API in JDK 9+ would then be build using -release 9 option. At the end the multi-release JAR would be build which would use the correct classes for the platform on which it is running. Are you OK with that approach?

> That would be a really nice idea to support other Java versions! You see, before the client was separated from TiSpark, we require Java 8 so that users do not struggle with incompatible Java versions.

So are you OK with approach to require JDK9+ (or even JDK 11) for building client-java while the produced jar would still work on JDK 8 ?

> I think there is always a way to build with JDK8 along with other Java versions... and it does not break the current environment.
> Let's try to fulfill them both while supporting more Java versions:)

The problem is that a build produced with JDK 8 javac will:

- produce runtime warnings when run on JDK9...15
- fail on JDK16+

[...]

PR #155, opened 2021-03-28
Allow building on JDK9+ by getting rid of DirectByteBuffer dependency

plevart:
Hi,

This is my 1st attempt at fixing https://github.com/tikv/client-java/issues/154 .

The problem when building/running with JDK9+ is not `sun.misc.Unsafe`. It is still there up to JDK15, you just have to `--add-module=jdk.unsupported`. In JDK16 it is encapsulated (you have to also add `--add-exports=...`) while in JDK17 it might get removed. The problem are: `sun.misc.Cleaner` which was removed in JDK9 (moved to `jdk.internal.ref` package) and `sun.nio.ch.DirectBuffer` interface which got encapsulated in JDK9.

So instead of building alternative versions of classes such as `MemoryUtil` compiled with and for JDK9+ and packaged into a multi-release JAR, I opted for simpler approach. I got rid of DirectByteBuffer usage (mostly they were used in TiBlockColumnVector) and replaced them with HeapByteBuffers that don't need `sun.misc.Cleaner` as they are allocated on heap (with byte[] array as their backing memory). As it turns out, code got much simpler using just public ByteBuffer API. And it should be at least as fast or even faster since various accesses to data types in HeapByteBuffer are performed with Unsafe internally, but in a safe and encapsulated way. No more manipulation with addresses and clumsy releasing of off-heap memory.

The only change to the build procedure is the addition of Maven profile `jdk9plus` which is activated when building with JDK9+ and just adds a dependency with scope `provided` (which means it is not packaged into the resulting artifact). This is needed because in JDK9+ `javax.annotation.Generated` annotation which is present on protobuf generated classes is missing (was renamed to `javax.annotation.processing.Generated` in JDK9) so this additional dependency which is added only when building with JDK9+ allows the project to be successfully built with JDK9+ (I tested with JDK11).

There is only one place left where `sun.misc.Unsafe` is still used: in `org.tikv.common.util.FastByteComparisons.LexicographicalComparerHolder.UnsafeComparer`. But this is not critical since as I said, `sun.misc.Unsafe` is still there for compilation at least up to JDK15 and in the code it has a fallback to somewhat slower variant that doesn't use unsafe. There is a fast `java.util.Arrays#compare(byte[], byte[])` method in JDK9+ but that would need separate compilation with JDK9+ and packaging into multi-release JAR. So this is still open.

So, WDYT?

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -519,4 +519,23 @@
             </plugin>
         </plugins>
     </build>
+
+    <profiles>
+        <profile>
+            <id>jdk9plus</id>
+            <activation>
+                <!-- activated when building with JDK9+ -->
+                <jdk>!1.8</jdk>
+            </activation>
+            <dependencies>
+                <dependency>
+                    <groupId>javax.annotation</groupId>
+                    <artifactId>javax.annotation-api</artifactId>
+                    <version>1.3.2</version>
+                    <scope>provided</scope>
+                </dependency>
+            </dependencies>
+        </profile>
+    </profiles>
+
 </project>
--- a/src/main/java/org/tikv/common/columnar/TiBlockColumnVector.java
+++ b/src/main/java/org/tikv/common/columnar/TiBlockColumnVector.java
@@ -15,12 +15,15 @@
 
 package org.tikv.common.columnar;
 
-import static org.tikv.common.util.MemoryUtil.EMPTY_BYTE_BUFFER_DIRECT;
+import static org.tikv.common.util.MemoryUtil.EMPTY_BYTE_BUFFER;
 
 import java.math.BigDecimal;
 import java.nio.ByteBuffer;
 import java.nio.charset.StandardCharsets;
-import java.sql.Timestamp;
+import java.time.Instant;
+import java.time.ZoneId;
+import java.time.ZonedDateTime;
+import java.util.Objects;
 import org.joda.time.LocalDate;
 import org.tikv.common.columnar.datatypes.CHType;
 import org.tikv.common.types.AbstractDateTimeType;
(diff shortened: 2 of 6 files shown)


marsishandsome:
image

please add sign-off

marsishandsome:
@birdstorm we should modify the ci to add java9 building step.

birdstorm:
> @birdstorm we should modify the ci to add java9 building step.

sure

plevart:
> please add sign-off

Done.

birdstorm:
@plevart Nice work! I think we should hold a contributor meeting this week about this change. 

There is some work to check before merge: 

- [ ] `TiBlockColumnVector` should behave the same way as it does, its related logic lies in TiSpark. Perhaps we can first move this logic out of the client and add it in the future?
- [x] The current CI script should be tested so that we can make sure it does not break the build.
- [ ] Add test for JDK 9+ environment on CI. It can be manually tested, and integrated with CI as well.
- [ ] Discuss which release should this PR be included. For now, I suggest we include this in release-4.0. i.e., the next big release.

plevart:
> @plevart Nice work! I think we should hold a contributor meeting this week about this change.
> 
> There is some work to check before merge:
> 
>     * [ ]  `TiBlockColumnVector` should behave the same way as it does, its related logic lies in TiSpark. Perhaps we can first move this logic out of the client and add it in the future?

Do you mean to remove TiBlockColumnVector and related logic out of java client? So currently client does not need it? I can look at what needs to be "deleted"...

> 
>     * [x]  The current CI script should be tested so that we can make sure it does not break the build.
> 
>     * [ ]  Add test for JDK 9+ environment on CI. It can be manually tested, and integrated with CI later.
> 
>     * [ ]  Discuss which release should this PR be included. For now, I suggest we include this in release-4.0. i.e., the next big release.

RIght. If the change removes TiBlockColumnVector then it should probably go to next major release. Does TiBlockColumnVector serve any purpose in the client or is it just a left-over from extracting the tikv client from TiSpark?

birdstorm:
> Do you mean to remove TiBlockColumnVector and related logic out of java client?

AFAIK, the TiBlockColumnVector is only used for TiSpark for now, it implements a columnar vector so that TiSpark can use columnar APIs to read from TiKV. It may be also useful for future logic which needs to read columnar data. @marsishandsome PTAL, we have to confirm it.

IMHO, this PR is great except that it lacks unit tests, but I can add it later. I will review this PR again tomorrow.

How about we arrange a meeting via zoom for more details, say Wednesday? @plevart @marsishandsome @ilovesoup I can schedule it according to the calendar. @plevart Feel free to tell us if you are occupied.

plevart:
> How about we arrange a meeting via zoom for more details, say Wednesday? @plevart @marsishandsome @ilovesoup I can schedule it according to the calendar. @plevart Feel free to tell us if you are occupied.

Wednesday is ok with me. Just note that I live in the GMT+2 time zone and would therefore appreciate the hour to be between say 10am to 8pm local time which is between 08:00 - 18:00 UTC if possible. Thanks.

birdstorm:
@plevart HI, after our discussion, we think this PR has resolved some concerns that:
1. it uses the profile that activates only when using JDK 9+, so the build process will be the same as it was.
2. the unsafe comparator for byte arrays is not changed.
3. its changes to `TiBlockColumnVector` will be discussed later because the author is busy now.

Overall, I believe there isn't much for us to discuss or share for now. If you have some more questions to ask, we can still schedule this meeting for tomorrow, since the zoom meeting is a bit heavy for just an elegant PR like this.

plevart:
I'm ok with waiting for the author of `TiBlockColumnVector` to chime in first. No need to discuss anything until then.

birdstorm:
Hi! @plevart Sorry to have kept you waiting.

We have decided to first check the correctness of replacing `DirectByteBuffer` with `HeapByteBuffer` in TiSpark. It would take a day or two, and estimated to have a desiring result by next Monday. After that, I will review this PR again, and it should be ready to merge by then.

birdstorm:
Hi! @plevart Sorry to have kept you waiting.

We have decided to first check the correctness of replacing `DirectByteBuffer` with `HeapByteBuffer` in TiSpark. It would take a day or two, and estimated to have a desiring result by next Monday. After that, I will review this PR again, and it should be ready to merge by then.

birdstorm:
Hi, @plevart FYI, the verify tests for `DirectByteBuffer` are undergoing in https://github.com/pingcap/tispark/pull/2005. You may track it for its progress.

birdstorm (approved):
LGTM

birdstorm:
In theory, `DirectByteBuffer` should cost less time when allocating than `HeapByteBuffer`. I will also benchmark it again when the `TiBlockColumnVector` is used in a large table to see if there are performance drawbacks.

plevart:
> In theory, `DirectByteBuffer` should cost less time when allocating than `HeapByteBuffer`. I will also benchmark it again when the `TiBlockColumnVector` is used in a large table to see if there are performance drawbacks.

This might be a common misconception - that native code can do allocations of "native" memory faster than java code can do heap allocations. Depending on the size of allocated object, java heap allocation translates into JIT-ed code as a mere increment of a pointer - by utilizing TLAB (thread-local allocation buffers) so it can be as fast as possible:

https://shipilev.net/jvm/anatomy-quarks/4-tlab-allocation/

But do benchmark the code. It's the only sure way to find out.

plevart:
DirectByteBuffer's purpose is mainly to interface native code where large amounts of data have to be exchanged between native code and java code. If consumers and producers of data are both pure-java-based, then heap ByteBuffer(s) are always a better choice.

PR #160, opened 2021-03-30
Remove unnecessary list copy

birdstorm:

There was an un-necessary list copy in groupKeysByRegion api.

remove it.

### Check List 

Tests 

 - Unit test
 - Integration test

Related changes

### DIFF
--- a/src/main/java/org/tikv/common/util/ClientUtils.java
+++ b/src/main/java/org/tikv/common/util/ClientUtils.java
@@ -121,7 +121,7 @@ public class ClientUtils {
 
   public static Map<TiRegion, List<ByteString>> groupKeysByRegion(
       RegionManager regionManager, List<ByteString> keys, BackOffer backoffer) {
-    return groupKeysByRegion(regionManager, new ArrayList<>(keys), backoffer, false);
+    return groupKeysByRegion(regionManager, keys, backoffer, false);
   }
 
   /**


src/main/java/org/tikv/common/util/ClientUtils.java
```diff
@@ -121,7 +121,7 @@ public static void appendBatches(
 
   public static Map<TiRegion, List<ByteString>> groupKeysByRegion(
       RegionManager regionManager, List<ByteString> keys, BackOffer backoffer) {
-    return groupKeysByRegion(regionManager, new ArrayList<>(keys), backoffer, false);
+    return groupKeysByRegion(regionManager, keys, backoffer, false);
```
marsishandsome, line 124: `keys` is sorted in line 137.
copy is necessary here to avoid multi-threads modification to the same keys List.

marsishandsome (approved):

src/main/java/org/tikv/common/util/ClientUtils.java
birdstorm: in which condition do we need to call groupKeysByRegion in concurrently?

marsishandsome: ok, let's remove it.
line 119 also do copy.

birdstorm: great!

PR #163, opened 2021-04-09
fix oom cause missing data bug

marsishandsome:
https://github.com/pingcap/tispark/pull/2010

oom during calling `client.coprocess` will cause missing data bug!

should not set `eof = true` if oom occures during calling `client.coprocess`

### Check List 

Tests 

 - Unit test
 - Integration test
 - No code

Code changes

 - Has persistent data change

Side effects

 - Increased code complexity

Related changes


birdstorm (approved):
LGTM

PR #174, opened 2021-05-25
Add CDCClient to client-java with a few tiny changes for TiFlink

shanzi:
This PR adds the pure java implementation of CDCClient which comunicates directly with TiKV with GRPC for change logs. It also includes some tiny changes to make TiFlink's implementation easier.
For details about TiFlink, please see [this link](https://internals.tidb.io/t/topic/124).

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
 
     <groupId>org.tikv</groupId>
     <artifactId>tikv-client-java</artifactId>
-    <version>3.2.0-SNAPSHOT</version>
+    <version>3.1.0-SNAPSHOT</version>
     <packaging>jar</packaging>
     <name>TiKV Java Client</name>
     <description>A Java Client for TiKV</description>
--- a/pom.xml
+++ b/pom.xml
@@ -337,8 +337,8 @@
                 <artifactId>maven-compiler-plugin</artifactId>
                 <version>3.7.0</version>
                 <configuration>
-                    <source>9</source>
-                    <target>9</target>
+                    <source>1.8</source>
+                    <target>1.8</target>
                     <encoding>UTF-8</encoding>
                     <showWarnings>true</showWarnings>
                     <showDeprecation>true</showDeprecation>
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
 
     <groupId>org.tikv</groupId>
     <artifactId>tikv-client-java</artifactId>
-    <version>3.1.0-SNAPSHOT</version>
+    <version>3.2.0-SNAPSHOT</version>
     <packaging>jar</packaging>
     <name>TiKV Java Client</name>
     <description>A Java Client for TiKV</description>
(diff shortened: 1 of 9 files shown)


marsishandsome:
build failed

```
[2021-05-25T13:36:04.891Z] [ERROR] /home/jenkins/agent/git/client-java/src/main/java/org/tikv/cdc/RegionCDCClient.java:[27,10] cannot find symbol

[2021-05-25T13:36:04.891Z] [ERROR]   symbol:   method of(org.tikv.kvproto.Cdcpb.Event.LogType,org.tikv.kvproto.Cdcpb.Event.LogType,org.tikv.kvproto.Cdcpb.Event.LogType,org.tikv.kvproto.Cdcpb.Event.LogType)

[2021-05-25T13:36:04.891Z] [ERROR]   location: interface java.util.Set
```

shanzi:
@marsishandsome What's the desire JDK version for java-client now? It's a java11 method.

marsishandsome:
> @marsishandsome What's the desire JDK version for java-client now? It's a java11 method.

java8

shanzi:
@marsishandsome Updated to java8

pom.xml
```diff
@@ -5,7 +5,7 @@
 
     <groupId>org.tikv</groupId>
     <artifactId>tikv-client-java</artifactId>
-    <version>3.1.0-SNAPSHOT</version>
+    <version>3.2.0-SNAPSHOT</version>
```
marsishandsome, line 8: `3.1.0` is not released yet, so we do not need to update version to `3.2.0`.

marsishandsome (approved):
LGTM

marsishandsome:
@birdstorm PTAL

birdstorm (approved):
LGTM

ti-srebot (approved):

birdstorm:
One more question, @shanzi can you please describe how to use multiple CDCClient at the same time? e.g., are there any precautions, and what would happen if the key ranges collide?

shanzi:
@birdstorm one CDCClient can only accept one key range, when using multiple CDCClients at the same time, each one will just work independently. The user is responsible for assigning proper key ranges for each. Thus overlapping key range won't have any problem. In other words, if the user runs two clients with overlapping key range, there will be two RegionCDCClient to the same region/store (That's why we don't use any Channel cache).

@marsishandsome Can I merge this now? And what's the schedule to release 3.1.0?

Thanks.

birdstorm:
> @birdstorm one CDCClient can only accept one key range, when using multiple CDCClients at the same time, each one will just work independently. The user is responsible for assigning proper key ranges for each. Thus overlapping key range won't have any problem. In other words, if the user runs two clients with overlapping key range, there will be two RegionCDCClient to the same region/store (That's why we don't use any Channel cache).
> 
> @marsishandsome Can I merge this now? And what's the schedule to release 3.1.0?
> 
> Thanks.
> 
> 

Will two RegionCDCClient to the same region/store be a problem? That is what I am concerned about.

shanzi:
@birdstorm There will be two channels to the single store. So, I think there won't be any issue.

Issue #179, opened 2021-06-03
Reduce the wait time after PD leader is switched


marsishandsome:
After the PD leader is switched, currently the client needs to wait 60 seconds, which is not necessary.

We should reduce the wait time after PD leader is switched.

PR #182, opened 2021-06-04
Fix pd retry not recovered for 60s

birdstorm:

close https://github.com/tikv/client-java/issues/179

When PD request is retried when leader is down, the leader info is not updated, so it will fail for at least 60s until the leader info is correct.


marsishandsome (approved):
LGTM

PR #192, opened 2021-06-11
Add CompareAndSet for RawClient and make Get returns Optional

andylokandy:

- Removed `putIfAbsent`
- Added `CompareAndSet`
- Make Get returns Optional

### Check List 

Tests 

 - Unit test
 - Integration test

Code changes

Related changes


src/main/java/org/tikv/raw/RawKVClient.java
```diff
@@ -135,37 +136,34 @@ public void put(ByteString key, ByteString value, long ttl) {
...
+   * Put a key-value pair if the prevValue matched the value in TiKV. This API is atomic.
    *
    * @param key key
    * @param value value
-   * @return a ByteString. returns ByteString.EMPTY if the value is written successfully. returns
-   *     the previous key if the value already exists, and does not write to TiKV.
    */
-  public ByteString putIfAbsent(ByteString key, ByteString value) {
-    return putIfAbsent(key, value, 0L);
+  public void compareAndSet(ByteString key, Optional<ByteString> prevValue, ByteString value) {
```
marsishandsome, line 144:
```suggestion
  public void compareAndSet(ByteString key, Optional<ByteString> prevValue, ByteString value)  throws RawCASConflictException {
```

src/main/java/org/tikv/common/exception/RawCASConflictException.java
```diff
@@ -0,0 +1,50 @@
+/*
+ * Copyright 2020 PingCAP, Inc.
```
marsishandsome, line 2:
```suggestion
 * Copyright 2021 PingCAP, Inc.
```

src/test/java/org/tikv/raw/RawKVClientTest.java
```diff
@@ -102,17 +103,12 @@ public void atomicAPITest() {
...
     try {
-      Thread.sleep(ttl * 1000);
-    } catch (InterruptedException e) {
-      Thread.currentThread().interrupt();
+      client.compareAndSet(key, null, value2, ttl);
+    } catch (RawCASConflictException err) {
+      assert err.getCurrValue() == Optional.of(value);
     }
-    ByteString res3 = client.putIfAbsent(key, value, ttl);
-    assert res3.isEmpty();
```
marsishandsome, line 115: can we keep the `ttl` test here?

marsishandsome (commented):
@birdstorm @Little-Wallace this PR changes the public API. I'm wondering whether the compatibility problem will affect users?

iosmanthus:
@andylokandy I don't think removing `putIfAbsent` is a good idea. What if the semantic of "key does not exist" changes?

src/main/java/org/tikv/common/region/RegionStoreClient.java
```diff
@@ -943,8 +947,13 @@ private void rawPutHelper(RawPutResponse resp) {
     }
   }
 
-  public ByteString rawPutIfAbsent(
```
marsishandsome, line 946: maybe we should keep this function for convenience.

andylokandy: `rawPutIfAbsent` acts very different with `CompareAndSet` on `prevValue` mismatch. The former will return the existing value while the latter throws an exception. I removed `rawPutIfAbsent` because I don't want to keep two different ways to notify CAS conflict.

src/test/java/org/tikv/raw/RawKVClientTest.java
andylokandy: First, the test is ignored. Second, this would never pass.

marsishandsome: @birdstorm PTAL

src/main/java/org/tikv/common/region/RegionStoreClient.java
marsishandsome: As I know one of our customers is using `rawPutIfAbsent`.  If this function is removed, there may be compatibility issues. @birdstorm @Little-Wallace

birdstorm: > As I know one of our customers is using `rawPutIfAbsent`.  If this function is removed, there may be compatibility issues. @birdstorm @Little-Wallace 

Right.

andylokandy:
@birdstorm @marsishandsome PTAL

src/main/java/org/tikv/raw/RawKVClient.java
```diff
@@ -152,20 +153,48 @@ public ByteString putIfAbsent(ByteString key, ByteString value) {
...
    * @param value value
    * @param ttl TTL of key (in seconds), 0 means the key will never be outdated.
-   * @return a ByteString. returns ByteString.EMPTY if the value is written successfully. returns
-   *     the previous key if the value already exists, and does not write to TiKV.
+   * @return a ByteString. returns Optional.EMPTY if the value is written successfully. returns the
+   *     previous key if the value already exists, and does not write to TiKV.
    */
-  public ByteString putIfAbsent(ByteString key, ByteString value, long ttl) {
-    String label = "client_raw_put_if_absent";
+  public Optional<ByteString> putIfAbsent(ByteString key, ByteString value, long ttl) {
```
marsishandsome, line 159: @birdstorm can the customer accept this API change?

src/test/java/org/tikv/raw/RawKVClientTest.java
birdstorm: In design, this test should not be ignored if we could determine that the tikv version is 5.0 and above. I suggest to leave it as it was.

.settings/org.eclipse.core.resources.prefs
```diff
@@ -0,0 +1,9 @@
+eclipse.preferences.version=1
```
birdstorm, line 1: These configurations for editor should not be included.

src/main/java/org/tikv/raw/RawKVClient.java
birdstorm: That is not very easy I think. We need to document all the changes so that they can update their code.

marsishandsome (approved):
LGTM

marsishandsome:
@birdstorm PTAL

birdstorm (approved):
LGTM

ti-srebot (approved):

marsishandsome:
@andylokandy test failed
```
[2021-06-17T02:31:49.545Z] [ERROR] /home/jenkins/agent/git/client-java/src/test/java/org/tikv/raw/RawKVClientTest.java:[123,41] incompatible types: java.util.Optional<com.google.protobuf.ByteString> cannot be converted to com.google.protobuf.ByteString

[2021-06-17T02:31:49.545Z] [ERROR] /home/jenkins/agent/git/client-java/src/test/java/org/tikv/raw/RawKVClientTest.java:[125,41] incompatible types: java.util.Optional<com.google.protobuf.ByteString> cannot be converted to com.google.protobuf.ByteString

[2021-06-17T02:31:49.545Z] [ERROR] /home/jenkins/agent/git/client-java/src/test/java/org/tikv/raw/RawKVClientTest.java:[132,41] incompatible types: java.util.Optional<com.google.protobuf.ByteString> cannot be converted to com.google.protobuf.ByteString
```

marsishandsome:
@andylokandy why use  the funciton name `compareAndSet` instead of `compareAndSwap` like the rust client does?

src/main/java/org/tikv/common/exception/RawCASConflictException.java
```diff
@@ -0,0 +1,50 @@
...
+  private final ByteString key;
+  private final Optional<ByteString> expectedPrevValue;
+  private final Optional<ByteString> prevValue;
+
+  public RawCASConflictException(
+      ByteString key, Optional<ByteString> expectedPrevValue, Optional<ByteString> prevValue) {
+    super(
+        String.format(
+            "key=%s expectedPrevValue=%s prevValue=%s",
+            KeyUtils.formatBytes(key), expectedPrevValue, prevValue));
```
marsishandsome, line 33: better to print the content of `expectedPrevValue` and `prevValue`.

currently 
```
org.tikv.common.exception.RawCASConflictException: key=72,101,108,108,111 expectedPrevValue=Optional[<ByteString@c9d0d6 size=3>] prevValue=Optional[<ByteString@6ccdb29f size=3>]

```

src/main/java/org/tikv/common/region/RegionStoreClient.java
```diff
@@ -964,13 +974,15 @@ public ByteString rawPutIfAbsent(
               regionManager, this, resp -> resp.hasRegionError() ? resp.getRegionError() : null);
       RawCASResponse resp =
           callWithRetry(backOffer, TikvGrpc.getRawCompareAndSwapMethod(), factory, handler);
-      return rawPutIfAbsentHelper(resp);
+      rawCompareAndSetHelper(key, prevValue, resp);
```
marsishandsome, line 977:
```suggestion
      return rawCompareAndSetHelper(key, prevValue, resp);
```

src/main/java/org/tikv/raw/RawKVClient.java
```diff
@@ -152,20 +153,48 @@ public ByteString putIfAbsent(ByteString key, ByteString value) {
...
+      throws RawCASConflictException {
+    String label = "client_raw_compare_and_swap";
     Histogram.Timer requestTimer = RAW_REQUEST_LATENCY.labels(label).startTimer();
     try {
       BackOffer backOffer = defaultBackOff();
       while (true) {
         RegionStoreClient client = clientBuilder.build(key);
         try {
-          ByteString result = client.rawPutIfAbsent(backOffer, key, value, ttl);
+          client.rawCompareAndSet(backOffer, key, prevValue, value, ttl);
```
marsishandsome, line 196: missing return

src/main/java/org/tikv/common/exception/RawCASConflictException.java
tisonkun:
```suggestion
            KeyUtils.formatBytes(key),
            KeyUtils.formatBytes(expectedPrevValue.orElse(ByteString.EMPTY)),
            KeyUtils.formatBytes(prevValue.orElse(ByteString.EMPTY))));
```

src/test/java/org/tikv/raw/RawKVClientTest.java
```diff
@@ -95,7 +96,24 @@ public void tearDown() throws Exception {
...
+    ByteString value = ByteString.copyFromUtf8("value");
+    ByteString value2 = ByteString.copyFromUtf8("value2");
+    client.delete(key);
+    client.compareAndSet(key, null, value, ttl);
+    try {
+      client.compareAndSet(key, null, value2, ttl);
+    } catch (RawCASConflictException err) {
+      assert err.getPrevValue() == Optional.of(value);
+    }
+  }
```
tisonkun, line 112:
```suggestion
  @Test
  public void rawCASTest() {
    if (!initialized) return;
    long ttl = 10;
    ByteString key = ByteString.copyFromUtf8("key_atomic");
    ByteString value = ByteString.copyFromUtf8("value");
    ByteString value2 = ByteString.copyFromUtf8("value2");
    client.delete(key);
    client.compareAndSet(key, Optional.empty(), value, ttl);
    Assert.assertEquals(value, client.get(key).get());
    try {
      client.compareAndSet(key, Optional.empty(), value2, ttl);
      Assert.fail("compareAndSet should fail.");
    } catch (RawCASConflictException err) {
      Assert.assertEquals(value, err.getPrevValue().get());
    }
  }
```

We'd better make use of `org.junit.Assert`. At least for new test cases.

tisonkun (commented):
Thanks for your updates @andylokandy ! I left an inline comment on testing part. PTAL.

src/test/java/org/tikv/raw/RawKVClientTest.java
```diff
@@ -95,7 +96,24 @@ public void tearDown() throws Exception {
 
   // tikv-4.0 does not support atomic api
   @Ignore
```
marsishandsome, line 98: We can enable this test now, by
- set default test tikv version from `relese-4.0` to `release-5.0`
- add tikv config: 
```
[storage]
enable-ttl = true
```

andylokandy:
@birdstorm @tisonkun @marsishandsome PTAL

src/main/java/org/tikv/common/exception/RawCASConflictException.java
tisonkun: Is this suggestion unapplied?

tisonkun (approved):
Generally LGTM. Thanks for your effort!

You may miss the comment https://github.com/tikv/client-java/pull/192/files#r653291717 .

tisonkun:
https://ci.pingcap.net/blue/organizations/jenkins/tikv-client-java_ghpr_integration_test/detail/tikv-client-java_ghpr_integration_test/41/pipeline

@andylokandy @marsishandsome test failed. Is it due to the test infra settings not upgrade?

marsishandsome:
> https://ci.pingcap.net/blue/organizations/jenkins/tikv-client-java_ghpr_integration_test/detail/tikv-client-java_ghpr_integration_test/41/pipeline
> 
> @andylokandy @marsishandsome test failed. Is it due to the test infra settings not upgrade?

cas is supported in `v5.0`, but current test uses `v4.0`.
please rebase this pr https://github.com/tikv/client-java/pull/202

PR #198, opened 2021-06-16
Support grpc forward

Little-Wallace:
When TiKV cluster deploy in three IDC A/B/C, if there is network isolation between the A and B, the message which should be send to leader in A can be forward by B.

### Check List 

Tests 

 - Unit test
 - Integration test
 - No code

Code changes

 - Has persistent data change

Side effects

 - Increased code complexity

Related changes

### DIFF
--- a/src/main/java/org/tikv/common/ConfigUtils.java
+++ b/src/main/java/org/tikv/common/ConfigUtils.java
@@ -48,6 +48,7 @@ public class ConfigUtils {
   public static final String TIKV_METRICS_PORT = "tikv.metrics.port";
 
   public static final String TIKV_NETWORK_MAPPING_NAME = "tikv.network.mapping";
+  public static final String TIKV_ENABLE_GRPC_FORWARD = "tikv.enable_grpc_forward";
 
   public static final String DEF_PD_ADDRESSES = "127.0.0.1:2379";
   public static final String DEF_TIMEOUT = "600ms";
--- a/src/main/java/org/tikv/common/ConfigUtils.java
+++ b/src/main/java/org/tikv/common/ConfigUtils.java
@@ -49,10 +49,12 @@ public class ConfigUtils {
 
   public static final String TIKV_NETWORK_MAPPING_NAME = "tikv.network.mapping";
   public static final String TIKV_ENABLE_GRPC_FORWARD = "tikv.enable_grpc_forward";
+  public static final String TIKV_GRPC_HEALTH_CHECK_TIMEOUT = "tikv.grpc.health_check_timeout";
 
   public static final String DEF_PD_ADDRESSES = "127.0.0.1:2379";
   public static final String DEF_TIMEOUT = "600ms";
   public static final String DEF_SCAN_TIMEOUT = "20s";
+  public static final int DEF_CHECK_HEALTH_TIMEOUT = 40;
   public static final int DEF_SCAN_BATCH_SIZE = 10240;
   public static final int DEF_MAX_FRAME_SIZE = 268435456 * 2; // 256 * 2 MB
   public static final int DEF_INDEX_SCAN_BATCH_SIZE = 20000;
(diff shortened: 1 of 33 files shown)


src/main/java/org/tikv/common/TiConfiguration.java
```diff
@@ -32,6 +33,8 @@
 
   private static final Logger logger = LoggerFactory.getLogger(TiConfiguration.class);
   private static final ConcurrentHashMap<String, String> settings = new ConcurrentHashMap<>();
+  public static final Metadata.Key FORWARD_META_DATA_KEY =
```
MyonKeminta, line 36:
```suggestion
  public static final Metadata.Key FORWARD_METADATA_KEY =
```

birdstorm (approved):
LGTM

src/main/java/org/tikv/common/ConfigUtils.java
```diff
@@ -76,6 +77,7 @@
   public static final boolean DEF_METRICS_ENABLE = false;
   public static final int DEF_METRICS_PORT = 3140;
   public static final String DEF_TIKV_NETWORK_MAPPING_NAME = "";
+  public static final boolean DEF_GRPC_FORWARD_ENABLE = true;
```
MyonKeminta, line 80: Why enable by default?

src/main/java/org/tikv/common/region/UnreachableStoreChecker.java
```diff
@@ -0,0 +1,77 @@
...
+
+  public void scheduleStoreHealthCheck(TiStore store) {
+    TiStore oldStore = this.stores.get(Long.valueOf(store.getId()));
+    if (oldStore != null) {
+      return;
+    }
+    synchronized (this.taskQueue) {
+      this.stores.put(Long.valueOf(store.getId()), store);
+      this.taskQueue.add(store);
+    }
```
MyonKeminta, line 35: Is this function possible to be called concurrently? May it be a race that multiple threads do the check-and-insert-if-not-exist operation at the same time?

src/main/java/org/tikv/common/region/UnreachableStoreChecker.java
```diff
@@ -0,0 +1,77 @@
...
+      HealthCheckRequest req = HealthCheckRequest.newBuilder().build();
+      try {
+        HealthCheckResponse resp = stub.check(req);
+        if (resp.getStatus() == HealthCheckResponse.ServingStatus.SERVING) {
+          store.markReachable();
+          this.stores.remove(Long.valueOf(store.getId()));
+          continue;
+        }
+      } finally {
+      }
```
MyonKeminta, line 63: What's the purpose of the try-finally?

src/main/java/org/tikv/common/region/UnreachableStoreChecker.java
```diff
@@ -0,0 +1,77 @@
...
+      return unhealthStore;
+    }
+  }
+
+  @Override
+  public Object call() throws Exception {
+    List<TiStore> unhealthStore = getUnhealthStore();
+    List<TiStore> restStore = new LinkedList<>();
+    for (TiStore store : unhealthStore) {
+      String addressStr = store.getStore().getAddress();
```
MyonKeminta, line 51: Did you check if the store becomes tombstone?

MyonKeminta (commented):
Rest LGTM, but I'm not very confident. Please confirm it works by tests.

src/main/java/org/tikv/common/ConfigUtils.java
Little-Wallace: why not?

MyonKeminta (approved):

ti-srebot (approved):

src/main/java/org/tikv/common/region/UnreachableStoreChecker.java
Little-Wallace: No, the store will not be tombstone after it was created. Because it is a final object.

PR #199, opened 2021-06-17
Fix scan exception when the start key is empty

### LINKED ISSUE #187: [BUG] raw kv scan cannot start with empty start key
**Describe the bug**

In ScanIterator.java, the start key cannot be empty.
```java
if (startKey.isEmpty()) {
  throw new IllegalArgumentException("start key cannot be empty");
} 
```
thus if RawKV wants to scan from the very beginning of all data, the ScanIterator will throw an exception.

**Java Client and TiDB/TiKV version info**

master

birdstorm:

Close #187 

Change the check condition for the scan operations to allow empty start key.

### Check List 

Tests 

 - Unit test
 - Integration test

Related changes

### DIFF
--- a/src/main/java/org/tikv/common/operation/iterator/ScanIterator.java
+++ b/src/main/java/org/tikv/common/operation/iterator/ScanIterator.java
@@ -50,11 +50,8 @@ public abstract class ScanIterator implements Iterator<Kvrpcpb.KvPair> {
       int limit,
       boolean keyOnly) {
     this.startKey = requireNonNull(startKey, "start key is null");
-    if (startKey.isEmpty()) {
-      throw new IllegalArgumentException("start key cannot be empty");
-    }
     this.endKey = Key.toRawKey(requireNonNull(endKey, "end key is null"));
-    this.hasEndKey = !endKey.equals(ByteString.EMPTY);
+    this.hasEndKey = !endKey.isEmpty();
     this.limit = limit;
     this.keyOnly = keyOnly;
     this.conf = conf;
@@ -74,7 +71,7 @@ public abstract class ScanIterator implements Iterator<Kvrpcpb.KvPair> {
     if (endOfScan || processingLastBatch) {
       return true;
     }
-    if (startKey == null || startKey.isEmpty()) {
+    if (startKey == null) {
       return true;
     }
     try {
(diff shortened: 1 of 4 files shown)


src/main/java/org/tikv/common/operation/iterator/ScanIterator.java
```diff
@@ -107,7 +104,7 @@ boolean cacheLoadFails() {
         startKey = lastKey.next().toByteString();
       }
       // notify last batch if lastKey is greater than or equal to endKey
-      if (hasEndKey && lastKey.compareTo(endKey) >= 0) {
+      if (hasEndKey && lastKey.compareTo(endKey) >= 0 || startKey.isEmpty()) {
         processingLastBatch = true;
         startKey = null;
       }
```
marsishandsome, line 110:
```suggestion
      if(startKey.isEmpty()) {
          processingLastBatch = true;
      }
      
      if (hasEndKey && lastKey.compareTo(endKey) >= 0) {
        processingLastBatch = true;
        startKey = null;
      }
```

src/main/java/org/tikv/common/operation/iterator/ScanIterator.java
```diff
@@ -107,7 +104,7 @@ boolean cacheLoadFails() {
         startKey = lastKey.next().toByteString();
       }
       // notify last batch if lastKey is greater than or equal to endKey
-      if (hasEndKey && lastKey.compareTo(endKey) >= 0) {
+      if (hasEndKey && lastKey.compareTo(endKey) >= 0 || startKey.isEmpty()) {
```
marsishandsome, line 107: `startKey.isEmpty()` always equals false

birdstorm: it is possible that startKey is ByteString.EMPTY, which indicates +♾

marsishandsome: ok, please update the comment

marsishandsome (approved):
LGTM

src/main/java/org/tikv/common/operation/iterator/ScanIterator.java
birdstorm: @marsishandsome done

Little-Wallace (approved):
LGTM

ti-srebot (approved):

Issue #205, opened 2021-06-22
[BUG] JVM does not exit after this PR https://github.com/tikv/client-java/pull/198


marsishandsome:
**Describe the bug**

JVM does not exit after this PR https://github.com/tikv/client-java/pull/198

**What did you do**

```
import org.tikv.common.TiConfiguration;
import org.tikv.common.TiSession;
import org.tikv.raw.RawKVClient;
import org.tikv.shade.com.google.protobuf.ByteString;

TiConfiguration conf = TiConfiguration.createRawDefault("127.0.0.1:2379");
TiSession session = TiSession.create(conf);
RawKVClient client = session.createRawClient();
// write (k1, v1) with ttl=30 seconds
client.put(ByteString.copyFromUtf8("k1"), ByteString.copyFromUtf8("v1"), 30);

// write (k2, v2) without ttl
client.put(ByteString.copyFromUtf8("k2"), ByteString.copyFromUtf8("v2"));

client.close();
session.close();
```

**What do you expect**

**What happens instead**

**Java Client and TiDB/TiKV version info**

PR #207, opened 2021-06-23
fix not stop scheduler

Little-Wallace:
Signed-off-by: Little-Wallace <bupt2013211450@gmail.com>

close https://github.com/tikv/client-java/issues/205

### Check List 

Tests 

 - Unit test
 - Integration test
 - No code

Code changes

 - Has persistent data change

Side effects

 - Increased code complexity

Related changes

### DIFF
--- a/src/main/java/org/tikv/common/TiSession.java
+++ b/src/main/java/org/tikv/common/TiSession.java
@@ -473,10 +473,8 @@ public class TiSession implements AutoCloseable {
     synchronized (sessionCachedMap) {
       sessionCachedMap.remove(conf.getPdAddrsString());
     }
-    synchronized (this) {
-      if (regionManager != null) {
-        regionManager.close();
-      }
+    if (regionManager != null) {
+      regionManager.close();
     }
     if (tableScanThreadPool != null) {
       tableScanThreadPool.shutdownNow();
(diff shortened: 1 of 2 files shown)


birdstorm (approved):
LGTM

src/main/java/org/tikv/common/TiSession.java
```diff
@@ -473,7 +473,11 @@ public synchronized void close() throws Exception {
     synchronized (sessionCachedMap) {
       sessionCachedMap.remove(conf.getPdAddrsString());
     }
-
+    synchronized (this) {
```
marsishandsome, line 476: line 461 already get the lock

src/main/java/org/tikv/common/TiSession.java
```diff
@@ -196,7 +196,7 @@ public Catalog getCatalog() {
     return res;
   }
 
-  public synchronized RegionManager getRegionManager() {
+  public RegionManager getRegionManager() {
```
marsishandsome, line 199: any reason to remove `synchronized` ?

Little-Wallace: Because it is safe to read the pointer concurrently. Any change for this pointer will be in synchronized.

Little-Wallace: Just like the method `getPDClient`

PR #215, opened 2021-06-24
Update document for v3.1.0

birdstorm:


README.md
```diff
@@ -80,7 +80,7 @@ public class Main {
 }
 ```
 
-### API
+### RawKVClient API
```
marsishandsome, line 83: maybe we should delete the api document here:
- 3.1 and 3.2 is different, users may confuse.
- we should deploy the java API document instead of list in README.

birdstorm: > maybe we should delete the api document here:
> 
> * 3.1 and 3.2 is different, users may confuse.
> * we should deploy the java API document instead of list in README.

Agree.

PR #241, opened 2021-07-26
fix backoff time

Little-Wallace:
Signed-off-by: Little-Wallace <bupt2013211450@gmail.com>

### Check List 

Tests 

 - Unit test
 - Integration test
 - No code

Code changes

 - Has persistent data change

Side effects

 - Increased code complexity

Related changes


src/main/java/org/tikv/common/operation/RegionErrorHandler.java
```diff
@@ -130,9 +130,11 @@ public boolean handleRegionError(BackOffer backOffer, Errorpb.Error error) {
           BackOffFunction.BackOffFuncType.BoServerBusy,
           new StatusRuntimeException(
               Status.fromCode(Status.Code.UNAVAILABLE).withDescription(error.toString())));
+      return true;
+    } else if (error.hasRegionNotFound()) {
```
birdstorm, line 134: region not found shall never happen

Little-Wallace: Why

marsishandsome (approved):
LGTM

PR #275, opened 2021-09-13
let class Key implmemnts Serializable

marsishandsome:

We create a additional class 
https://github.com/tikv/migration/blob/main/online-bulk-load/src/main/scala/org/tikv/bulkload/SerializableKey.scala, because Key does not implmemnt Serializable.

let class Key implmemnts Serializable, so that SerializableKey is no more needed.

### Check List 

Tests 

 - Unit test
 - Integration test
 - No code

Code changes

 - Has persistent data change

Side effects

 - Increased code complexity

Related changes


Little-Wallace (approved):
LGTM

birdstorm (approved):
LGTM

ti-srebot (approved):

PR #284, opened 2021-10-09
Support SST Decoder

marsishandsome:

close https://github.com/tikv/client-java/issues/285

support decode SST to Key-Value

subtask of https://github.com/tikv/tikv/issues/11007

RocksDB provides a java API [SstFileReader](https://github.com/facebook/rocksdb/blob/6.23.fb/java/src/main/java/org/rocksdb/SstFileReader.java) to decode SST to Key-Value pairs. We can wrapper this API in client-java.

### Check List 

Tests 

 - Unit test
 - Integration test
 - No code

Code changes

 - Has persistent data change

Side effects

 - Increased code complexity

Related changes


marsishandsome:
> I suggest checking [`SSTMeta.data_encode`](https://github.com/andylokandy/kvproto/blob/5cbe1a0b8ca53964341f7f802ce98dcd62ce1565/proto/import_sstpb.proto#L93). This should help avoid silent bugs after enabling API V2.

Cool. Can we do this job as a subtask of API V2?

PR #290, opened 2021-10-21
scan end key push down.

zekisong:
while do rawScan, end key not push down to the tikv region, so small scan will cause heavy load. 
by default every scan operation will fetch 10240 key values from region even though expect fetch zero result
fix it by setEndKey for RawScanRequest if if necessary


zz-jason:
could you add some tests?

zz-jason:
@zekisong I'm going to close this PR since it hasn't been updated for a long time. Feel free to reopen it once you wish to contribute to this PR again in the future. Thanks.

PR #306, opened 2021-11-03
Retry when OtherError contains message: region xx is missing

birdstorm:
Retry when OtherError contains message: `region xxx is missing`

`Region is missing` indicates the required region does not exist on the current store. It may have been migrated to other stores. We could safely retry the query because the region cache has been invalidated.


marsishandsome (commented):
If region is missing, we cannot retry in RegionStoreClient.
We should refetch the region info, and create a new RegionnStoreClient.

birdstorm:
> If region is missing, we cannot retry in RegionStoreClient. We should refetch the region info, and create a new RegionnStoreClient.

right

PR #316, opened 2021-11-05
Fix gRPC forward bug when store is unreachable

birdstorm:
- Fix grpc forward bug.
- Modify default backoff from 100ms to 20ms.


marsishandsome (approved):
LGTM

marsishandsome (approved):
LGTM

PR #325, opened 2021-11-14
Pr/324

Little-Wallace:
### Check List 

Tests 

 - Unit test
 - Integration test
 - No code

Code changes

 - Has persistent data change

Side effects

 - Increased code complexity

Related changes

### DIFF
--- a/src/main/java/org/tikv/common/region/AbstractRegionStoreClient.java
+++ b/src/main/java/org/tikv/common/region/AbstractRegionStoreClient.java
@@ -20,13 +20,18 @@ package org.tikv.common.region;
 import static com.google.common.base.Preconditions.checkArgument;
 import static com.google.common.base.Preconditions.checkNotNull;
 
+import com.google.common.util.concurrent.ListenableFuture;
 import com.google.protobuf.ByteString;
 import io.grpc.ManagedChannel;
 import io.grpc.Metadata;
+import io.grpc.health.v1.HealthCheckRequest;
+import io.grpc.health.v1.HealthCheckResponse;
+import io.grpc.health.v1.HealthGrpc;
 import io.grpc.stub.ClientCalls;
 import io.grpc.stub.MetadataUtils;
 
 import java.util.ArrayList;
+import java.util.LinkedList;
 import java.util.List;
 import java.util.Set;
 import java.util.concurrent.*;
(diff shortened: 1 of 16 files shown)


src/main/java/org/tikv/common/region/AbstractRegionStoreClient.java
```diff
@@ -110,215 +108,200 @@ public boolean onNotLeader(TiRegion newRegion) {
...
+              if (!resp.hasRegionError()) {
+                // the peer is leader
+                return task.peer;
+              }
+            }
+          } catch (Exception ignored) {
+          }
+        } else {
+          unfinished.add(task);
+        }
```
zz-jason, line 237:
```suggestion
        if (!task.task.isDone()) {
          unfinished.add(task);
          continue;
        }
        try {
          Kvrpcpb.RawGetResponse resp = task.task.get();
          if (!resp.hasRegionError()) {
            // the peer is leader
            return task.peer;
          }
        } catch (Exception ignored) {
        }
```

src/main/java/org/tikv/common/region/AbstractRegionStoreClient.java
```diff
@@ -110,215 +108,200 @@ public boolean onNotLeader(TiRegion newRegion) {
...
-    if (retryForwardTimes >= region.getFollowerList().size()) {
-      // If we try to forward request to leader by follower failed, it means that the store of old
-      // leader may be
-      // unavailable but the new leader has not been report to PD. So we can ban this store for a
-      // short time to
-      // avoid too many request try forward rather than try other peer.
-      if (originStore != null) {
-        originStore.forwardFail();
+    while (true) {
+      try {
```
zz-jason, line 217: could you add these metrics for each request:
- the total backoff time
	- the total backoff time caused by grpc forward
	- the total backoff time caused by switch leader
- the total backoff operation
	- the total backoff operation for grpc forward
	- the total backoff operation for switch leader

with these metrics, we can see where the time is spent, and the detailed reason.

src/main/java/org/tikv/common/region/AbstractRegionStoreClient.java
```diff
@@ -110,215 +108,200 @@ public boolean onNotLeader(TiRegion newRegion) {
...
-      if (originStore != null) {
-        originStore.forwardFail();
-      } else {
-        targetStore.forwardFail();
+      List<SwitchLeaderTask> unfinished = new LinkedList<>();
+      for (SwitchLeaderTask task : responses) {
+        if (task.task.isDone()) {
+          try {
+            Kvrpcpb.RawGetResponse resp = task.task.get();
+            if (resp != null) {
```
zz-jason, line 227: no need to check whether `resp` is null since you will catch all the exceptions happened in this code block

PR #333, opened 2021-11-18
TwoPhaseCommiter log should be debug level

humengyu2012:
The log of `org.tikv.txn.TwoPhaseCommitter` is too large,  which will cause the really useful logs to be hidden.

image


sunxiaoguang (approved):
LGTM

sunxiaoguang:
@humengyu2012 please sign DCO

PR #335, opened 2021-11-19
Optimize seek leader store logic

birdstorm:
remove redundant seek leader logic that may cause retry.


src/main/java/org/tikv/common/region/AbstractRegionStoreClient.java
```diff
@@ -240,6 +212,12 @@ private Boolean seekLeaderStore(SlowLog slowLog) {
           updateClientStub();
           return true;
         }
```
zz-jason, line 214: should we add a log here to indicate that the leader is unreachable?

marsishandsome (approved):
LGTM

PR #337, opened 2021-11-20
Fix pd request throws invalid store id

birdstorm:
When the Java Client requests a store id that PD does not recognize, e.g., the PD was scaled-in and does not know about a scaled-out TiKV, the PD will respond with a `StatusRuntimeException` rather than an error wrapped by Response itself.

Thus we should resolve StatusRuntimeException so that the invalid store info can be dealt with correctly by clearing all-region/store cache info.

### DIFF
--- a/src/main/java/org/tikv/common/operation/PDErrorHandler.java
+++ b/src/main/java/org/tikv/common/operation/PDErrorHandler.java
@@ -20,6 +20,8 @@ package org.tikv.common.operation;
 import static org.tikv.common.pd.PDError.buildFromPdpbError;
 
 import java.util.function.Function;
+
+import io.grpc.StatusRuntimeException;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.tikv.common.PDClient;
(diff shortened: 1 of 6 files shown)


src/main/java/org/tikv/common/operation/PDErrorHandler.java
```diff
@@ -74,6 +75,10 @@ public boolean handleResponseError(BackOffer backOffer, RespT resp) {
 
   @Override
   public boolean handleRequestError(BackOffer backOffer, Exception e) {
+    // store id is not found
+    if (e instanceof StatusRuntimeException && e.getMessage().contains("invalid store ID")) {
```
zz-jason, line 79: can we directly check whether the instance is of `InvalidStoreException`?

zz-jason: BTW, should the error message be `Invalid storeId`?

src/main/java/org/tikv/common/region/RegionManager.java
```diff
@@ -222,6 +222,15 @@ public TiStore getStoreById(long id) {
     return getStoreById(id, defaultBackOff());
   }
 
+  public TiStore getStoreById(long id, BackOffer backOffer) {
+    TiStore store = getStoreByIdWithBackOff(id, backOffer);
+    if (store == null) {
+      cache.clearAll();
```
zz-jason, line 228: we should log info or warn messages for this critical operation.

src/main/java/org/tikv/common/operation/PDErrorHandler.java
birdstorm: no, the exception tikv returns is a grpc exception, it is a StatusRuntimeException that only contains "invalid store id xx" information in its message.

zz-jason (approved):
LGTM

marsishandsome (dismissed):
LGTM

ti-srebot (dismissed):

marsishandsome (approved):
LGTM

PR #345, opened 2021-11-27
remove unnecessary isDebugEnabled() check

zz-jason:

make code clean, may reduce unnecessary branch predication

remove unnecessary `isDebugEnabled()` check

### Check List 

Tests 


marsishandsome (approved):
LGTM

sunxiaoguang:
String.format is slow, if log is not guarded by isDebugEnabled check, it's going to waste cycles unless they are only called rarely.

birdstorm (changes requested):
I don't think this works as expected. Debug should not be called rapidly because String.format is time-consuming.

zz-jason:
I see, is there any way to lazily compute the string?

sunxiaoguang:
> I see, is there any way to lazily compute the string?

slf4j has some overloaded functions to pass parameters in array. The log message is rendered only when log level satisfies therefore avoid unnecessary format. However that works only when the arguments are values that do not need to evaluate.

zz-jason:
TiDB optimized this by passing a stringer() function to the logger. The message is only evaluated when the log level is reached, no matter the value message is a string or a function call. It helps TiDB to reduce about 5% CPU overhead in high concurrency simply workloads.

I'm going to close this PR. We can refactor the logger system in the future when we have such a logger wrapper like TiDB.

PR #352, opened 2021-11-30
check timeout during SeekLeader and SeekProxy

marsishandsome:

close https://github.com/tikv/client-java/issues/353

we should check timeout during SeekLeader and SeekProxy

### DIFF
--- a/src/main/java/org/tikv/common/region/AbstractRegionStoreClient.java
+++ b/src/main/java/org/tikv/common/region/AbstractRegionStoreClient.java
@@ -35,9 +35,8 @@ import org.slf4j.LoggerFactory;
 import org.tikv.common.AbstractGRPCClient;
 import org.tikv.common.TiConfiguration;
 import org.tikv.common.exception.GrpcException;
-import org.tikv.common.log.SlowLog;
-import org.tikv.common.log.SlowLogEmptyImpl;
 import org.tikv.common.log.SlowLogSpan;
+import org.tikv.common.util.BackOffer;
 import org.tikv.common.util.ChannelFactory;
 import org.tikv.kvproto.Kvrpcpb;
 import org.tikv.kvproto.Metapb;
@@ -81,10 +80,6 @@ public abstract class AbstractRegionStoreClient
     this.store = store;
     if (this.store.getProxyStore() != null) {
       this.timeout = conf.getForwardTimeout();
-    } else if (!this.store.isReachable()) {
-      // cannot get Deadline or SlowLog instance here
-      // use SlowLogEmptyImpl instead to skip slow log record
-      onStoreUnreachable(SlowLogEmptyImpl.INSTANCE);
     }
   }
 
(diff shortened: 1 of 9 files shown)


src/test/java/org/tikv/raw/RawKVClientTest.java
```diff
@@ -204,6 +204,26 @@ public void testDeadlineBackOff() {
     }
   }
 
+  @Test
+  public void testBackoffTimeout() {
+    int timeout = 2000;
```
zz-jason, line 209: sleep 2 seconds will slow down the unit test, could you reduce the time to sleep to reduce the test time?

src/test/java/org/tikv/raw/RawKVClientTest.java
```diff
@@ -204,6 +204,26 @@ public void testDeadlineBackOff() {
...
+    try {
+      while (true) {
+        Thread.sleep(sleep);
+        backOffer.checkTimeout();
+      }
+    } catch (Exception ignored) {
+    } finally {
+      long e = System.currentTimeMillis();
+      long duration = e - s;
+      logger.info("duration = " + duration);
```
zz-jason, line 222: can we remove this log?

src/main/java/org/tikv/common/region/AbstractRegionStoreClient.java
```diff
@@ -81,10 +80,6 @@ protected AbstractRegionStoreClient(
     this.store = store;
     if (this.store.getProxyStore() != null) {
       this.timeout = conf.getForwardTimeout();
-    } else if (!this.store.isReachable()) {
```
zz-jason, line 84: why it's removed?

src/main/java/org/tikv/common/util/BackOffFunction.java
```diff
@@ -63,6 +63,7 @@ private int expo(int base, int cap, int n) {
     BoRegionMiss,
     BoUpdateLeader,
     BoServerBusy,
-    BoTxnNotFound
+    BoTxnNotFound,
+    CheckTimeout
```
zz-jason, line 67: how about `BoTimeout` to stay consistent with the existing enums?

src/test/java/org/tikv/raw/RawKVClientTest.java
```diff
@@ -204,6 +204,26 @@ public void testDeadlineBackOff() {
...
+      assert (duration <= timeout + sleep);
```
zz-jason, line 223: how about using JUnit's ` assertTrue`:

```java
assertTrue(duration <= timeout + sleep);
```

zz-jason:
@marsishandsome please remove the unnecessary items in the PR description.

birdstorm (dismissed):
LGTM

src/main/java/org/tikv/common/region/AbstractRegionStoreClient.java
marsishandsome: We do not need to do `Seek Leader` and `Seek Proxy` during initializing `RegionStoreClient`. Reasons are:
- `Seek Leader` and `Seek Proxy` cost time, but we do not want to the initializing process cost too much time.
- `Seek Leader` and `Seek Proxy` will happen after the first call when reciving gRPC Deadline Exceeded.

birdstorm (approved):
LGTM

zz-jason (approved):
LGTM

ti-srebot (approved):

zz-jason (approved):

PR #358, opened 2021-12-02
[close #372] add circuit breaker and smart rawkv client

### LINKED ISSUE #372: support circuit breaker and integrate it with a smart rawkv client
**Is your feature request related to a problem? Please describe.**

once the cluster is gone, we need to discover that issue, respond to errors to the users immediately.

**Describe the solution you'd like**

1. can discover that the TiKV cluster is gone, respond to errors to the users immediately.
2. can discover that the TiKV cluster is recovered, respond to users with correct results in a reasonable time

**Describe alternatives you've considered**

circuit breaker implemented in https://github.com/alibaba/Sentinel/ and https://github.com/Netflix/Hystrix/

**Additional context**

N/A

marsishandsome:

add Circuit Breaker

### DIFF
--- a/src/main/java/org/tikv/common/ConfigUtils.java
+++ b/src/main/java/org/tikv/common/ConfigUtils.java
@@ -92,17 +92,17 @@ public class ConfigUtils {
   public static final String TIKV_KEY_CERT_CHAIN = "tikv.key_cert_chain";
   public static final String TIKV_KEY_FILE = "tikv.key_file";
 
-  public static final String TiKV_CIRCUIT_BREAK_ENABLE = "tikv.circuit.break.enable";
+  public static final String TiKV_CIRCUIT_BREAK_ENABLE = "tikv.circuit_break.enable";
   public static final String TiKV_CIRCUIT_BREAK_AVAILABILITY_WINDOW_IN_SECONDS =
-      "tikv.circuit.break.trigger.availability.window_in_seconds";
+      "tikv.circuit_break.trigger.availability.window_in_seconds";
   public static final String TiKV_CIRCUIT_BREAK_AVAILABILITY_ERROR_THRESHOLD_PERCENTAGE =
-      "tikv.circuit.break.trigger.availability.error_threshold_percentage";
+      "tikv.circuit_break.trigger.availability.error_threshold_percentage";
   public static final String TiKV_CIRCUIT_BREAK_AVAILABILITY_REQUEST_VOLUMN_THRESHOLD =
-      "tikv.circuit.break.trigger.availability.request_volumn_threshold";
+      "tikv.circuit_break.trigger.availability.request_volumn_threshold";
   public static final String TiKV_CIRCUIT_BREAK_SLEEP_WINDOW_IN_SECONDS =
-      "tikv.circuit.break.trigger.sleep_window_in_seconds";
+      "tikv.circuit_break.trigger.sleep_window_in_seconds";
   public static final String TiKV_CIRCUIT_BREAK_ATTEMPT_REQUEST_COUNT =
-      "tikv.circuit.break.trigger.attempt_request_count";
+      "tikv.circuit_break.trigger.attempt_request_count";
 
   public static final String TIFLASH_ENABLE = "tiflash.enable";
   public static final String DEF_PD_ADDRESSES = "127.0.0.1:2379";
--- a/src/main/java/org/tikv/common/ConfigUtils.java
+++ b/src/main/java/org/tikv/common/ConfigUtils.java
@@ -92,6 +92,18 @@ public class ConfigUtils {
   public static final String TIKV_KEY_CERT_CHAIN = "tikv.key_cert_chain";
   public static final String TIKV_KEY_FILE = "tikv.key_file";
 
+  public static final String TiKV_CIRCUIT_BREAK_ENABLE = "tikv.circuit.break.enable";
+  public static final String TiKV_CIRCUIT_BREAK_AVAILABILITY_WINDOW_IN_SECONDS =
+      "tikv.circuit.break.trigger.availability.window_in_seconds";
+  public static final String TiKV_CIRCUIT_BREAK_AVAILABILITY_ERROR_THRESHOLD_PERCENTAGE =
+      "tikv.circuit.break.trigger.availability.error_threshold_percentage";
+  public static final String TiKV_CIRCUIT_BREAK_AVAILABILITY_REQUEST_VOLUMN_THRESHOLD =
+      "tikv.circuit.break.trigger.availability.request_volumn_threshold";
+  public static final String TiKV_CIRCUIT_BREAK_SLEEP_WINDOW_IN_SECONDS =
+      "tikv.circuit.break.trigger.sleep_window_in_seconds";
+  public static final String TiKV_CIRCUIT_BREAK_ATTEMPT_REQUEST_COUNT =
+      "tikv.circuit.break.trigger.attempt_request_count";
+
   public static final String TIFLASH_ENABLE = "tiflash.enable";
   public static final String DEF_PD_ADDRESSES = "127.0.0.1:2379";
   public static final String DEF_TIMEOUT = "200ms";
@@ -161,4 +173,11 @@ public class ConfigUtils {
   public static final int DEF_TIKV_GRPC_KEEPALIVE_TIMEOUT = 3;
   public static final boolean DEF_TIKV_TLS_ENABLE = false;
   public static final boolean DEF_TIFLASH_ENABLE = false;
+
+  public static final boolean DEF_TiKV_CIRCUIT_BREAK_ENABLE = false;
+  public static final int DEF_TiKV_CIRCUIT_BREAK_AVAILABILITY_WINDOW_IN_SECONDS = 60;
+  public static final int DEF_TiKV_CIRCUIT_BREAK_AVAILABILITY_ERROR_THRESHOLD_PERCENTAGE = 100;
+  public static final int DEF_TiKV_CIRCUIT_BREAK_AVAILABILITY_REQUST_VOLUMN_THRESHOLD = 10;
+  public static final int DEF_TiKV_CIRCUIT_BREAK_SLEEP_WINDOW_IN_SECONDS = 20;
+  public static final int DEF_TiKV_CIRCUIT_BREAK_ATTEMPT_REQUEST_COUNT = 10;
 }
--- a/src/main/java/org/tikv/common/failsafe/CircuitBreakerImpl.java
+++ b/src/main/java/org/tikv/common/failsafe/CircuitBreakerImpl.java
@@ -104,13 +104,11 @@ public class CircuitBreakerImpl implements CircuitBreaker {
     return !isOpen();
   }
 
-  @Override
-  public boolean isOpen() {
+  boolean isOpen() {
     return circuitOpened.get() >= 0;
   }
 
-  @Override
-  public Status getStatus() {
+  Status getStatus() {
     return status.get();
   }
 
--- a/src/main/java/org/tikv/common/failsafe/CircuitBreakerImpl.java
+++ b/src/main/java/org/tikv/common/failsafe/CircuitBreakerImpl.java
@@ -1,0 +1,187 @@
+/*
+ * Copyright 2021 PingCAP, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.tikv.common.failsafe;
+
+import java.util.concurrent.atomic.AtomicLong;
+import java.util.concurrent.atomic.AtomicReference;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.tikv.common.TiConfiguration;
+
+public class CircuitBreakerImpl implements CircuitBreaker {
+  private static final Logger logger = LoggerFactory.getLogger(CircuitBreakerImpl.class);
+
+  private final boolean enable;
+  private final int windowInSeconds;
+  private final int errorThresholdPercentage;
+  private final int requestVolumeThreshold;
+  private final int sleepWindowInSeconds;
+  private final int attemptRequestCount;
+
+  private final AtomicLong circuitOpened = new AtomicLong(-1);
+  private final AtomicReference<Status> status = new AtomicReference<>(Status.CLOSED);
+  private final AtomicLong attemptCount = new AtomicLong(0);
+  private final AtomicLong attemptSuccessCount = new AtomicLong(0);
+
+  private final CircuitBreakerMetrics metrics;
+
+  public CircuitBreakerImpl(TiConfiguration conf) {
+    this(
+        conf.isCircuitBreakEnable(),
+        conf.getCircuitBreakAvailabilityWindowInSeconds(),
+        conf.getCircuitBreakAvailabilityErrorThresholdPercentage(),
+        conf.getCircuitBreakAvailabilityRequestVolumnThreshold(),
+        conf.getCircuitBreakSleepWindowInSeconds(),
+        conf.getCircuitBreakAttemptRequestCount());
+  }
+
+  public CircuitBreakerImpl(
+      boolean enable,
+      int windowInSeconds,
+      int errorThresholdPercentage,
+      int requestVolumeThreshold,
+      int sleepWindowInSeconds,
+      int attemptRequestCount) {
+    this.enable = enable;
+    this.windowInSeconds = windowInSeconds;
+    this.errorThresholdPercentage = errorThresholdPercentage;
+    this.requestVolumeThreshold = requestVolumeThreshold;
+    this.sleepWindowInSeconds = sleepWindowInSeconds;
+    this.attemptRequestCount = attemptRequestCount;
+    this.metrics = new CircuitBreakerMetricsImpl(windowInSeconds);
+    this.metrics.addListener(getMetricsListener());
+  }
+
+  private MetricsListener getMetricsListener() {
+    return hc -> {
+      logger.info("onNext " + hc.toString());
+      // check if we are past the requestVolumeThreshold
+      if (hc.getTotalRequests() < requestVolumeThreshold) {
+        // we are not past the minimum volume threshold for the stat window,
+        // so no change to circuit status.
+        // if it was CLOSED, it stays CLOSED
+        // if it was half-open, we need to wait for some successful command executions
+        // if it was open, we need to wait for sleep window to elapse
+      } else {
+        if (hc.getErrorPercentage() < errorThresholdPercentage) {
+          // we are not past the minimum error threshold for the stat window,
+          // so no change to circuit status.
+          // if it was CLOSED, it stays CLOSED
+          // if it was half-open, we need to wait for some successful command executions
+          // if it was open, we need to wait for sleep window to elapse
+        } else {
+          // our failure rate is too high, we need to set the state to OPEN
+          close2Open();
+        }
+      }
+    };
+  }
+
+  @Override
+  public CircuitBreakerMetrics getMetrics() {
+    return metrics;
+  }
+
+  @Override
+  public boolean allowRequest() {
+    if (!enable) {
+      return true;
+    }
+    return !isOpen();
+  }
+
+  @Override
+  public boolean isOpen() {
+    return circuitOpened.get() >= 0;
+  }
+
+  @Override
+  public Status getStatus() {
+    return status.get();
+  }
+
+  @Override
+  public void markAttemptSuccess() {
+    if (attemptSuccessCount.incrementAndGet() >= this.attemptRequestCount) {
+      halfOpen2Close();
+    }
+  }
+
+  @Override
+  public void markAttemptFailure() {
+    halfOpen2Open();
+  }
+
+  @Override
+  public boolean attemptExecution() {
+    if (!isOpen()) {
+      return true;
+    } else {
+      if (isAfterSleepWindow()) {
+        // only the `attemptRequestCount` requests after sleep window should execute
+        // if all the executing commands succeed, the status will transition to CLOSED
+        // if some of the executing commands fail, the status will transition to OPEN
+        open2HalfOpen();
+        return attemptCount.incrementAndGet() <= attemptRequestCount;
+      } else {
+        return false;
+      }
+    }
+  }
+
+  private boolean isAfterSleepWindow() {
+    final long circuitOpenTime = circuitOpened.get();
+    final long currentTime = System.currentTimeMillis();
+    final long sleepWindowTime = (long) sleepWindowInSeconds * 1000;
+    return currentTime >= circuitOpenTime + sleepWindowTime;
+  }
+
+  private void close2Open() {
+    if (status.compareAndSet(Status.CLOSED, Status.OPEN)) {
+      // This thread wins the race to open the circuit
+      // it sets the start time for the sleep window
+      circuitOpened.set(System.currentTimeMillis());
+      logger.info("CLOSED => OPEN");
+    }
+  }
+
+  private void halfOpen2Close() {
+    if (status.compareAndSet(Status.HALF_OPEN, Status.CLOSED)) {
+      // This thread wins the race to close the circuit
+      circuitOpened.set(-1L);
+      logger.info("HALF_OPEN => CLOSED");
+    }
+  }
+
+  private void open2HalfOpen() {
+    if (status.compareAndSet(Status.OPEN, Status.HALF_OPEN)) {
+      // This thread wins the race to half close the circuit
+      // it resets the attempt count
+      attemptCount.set(0);
+      attemptSuccessCount.set(0);
+      logger.info("OPEN => HALF_OPEN");
+    }
+  }
+
+  private void halfOpen2Open() {
+    if (status.compareAndSet(Status.HALF_OPEN, Status.OPEN)) {
+      // This thread wins the race to re-open the circuit
+      // it resets the start time for the sleep window
+      circuitOpened.set(System.currentTimeMillis());
+      logger.info("HALF_OPEN => OPEN");
+    }
+  }
+}
--- a/src/main/java/org/tikv/common/failsafe/CircuitBreakerImpl.java
+++ b/src/main/java/org/tikv/common/failsafe/CircuitBreakerImpl.java
@@ -113,14 +113,14 @@ public class CircuitBreakerImpl implements CircuitBreaker {
   }
 
   @Override
-  public void markAttemptSuccess() {
+  public void recordAttemptSuccess() {
     if (attemptSuccessCount.incrementAndGet() >= this.attemptRequestCount) {
       halfOpen2Close();
     }
   }
 
   @Override
-  public void markAttemptFailure() {
+  public void recordAttemptFailure() {
     halfOpen2Open();
   }
 
--- a/src/main/java/org/tikv/common/failsafe/CircuitBreakerImpl.java
+++ b/src/main/java/org/tikv/common/failsafe/CircuitBreakerImpl.java
@@ -126,7 +126,7 @@ public class CircuitBreakerImpl implements CircuitBreaker {
 
   @Override
   public boolean attemptExecution() {
-    if (!isOpen()) {
+    if (allowRequest()) {
       return true;
     } else {
       if (isAfterSleepWindow()) {
--- a/src/main/java/org/tikv/common/failsafe/CircuitBreakerImpl.java
+++ b/src/main/java/org/tikv/common/failsafe/CircuitBreakerImpl.java
@@ -15,6 +15,7 @@
 
 package org.tikv.common.failsafe;
 
+import java.io.IOException;
 import java.util.concurrent.atomic.AtomicLong;
 import java.util.concurrent.atomic.AtomicReference;
 import org.slf4j.Logger;
@@ -183,5 +184,10 @@ public class CircuitBreakerImpl implements CircuitBreaker {
       circuitOpened.set(System.currentTimeMillis());
       logger.info("HALF_OPEN => OPEN");
     }
+  }
+
+  @Override
+  public void close() throws IOException {
+    metrics.close();
   }
 }
--- a/src/main/java/org/tikv/common/failsafe/CircuitBreakerMetricsImpl.java
+++ b/src/main/java/org/tikv/common/failsafe/CircuitBreakerMetricsImpl.java
@@ -56,13 +56,13 @@ public class CircuitBreakerMetricsImpl implements CircuitBreakerMetrics {
   }
 
   @Override
-  public void success() {
-    currentMetrics.get().success();
+  public void recordSuccess() {
+    currentMetrics.get().recordSuccess();
   }
 
   @Override
-  public void failure() {
-    currentMetrics.get().failure();
+  public void recordFailure() {
+    currentMetrics.get().recordFailure();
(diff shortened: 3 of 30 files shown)


src/main/java/org/tikv/common/failsafe/CircuitBreakerMetricsImpl.java
```diff
@@ -0,0 +1,91 @@
...
+  public CircuitBreakerMetricsImpl(int windowInSeconds) {
+    this.windowInMS = windowInSeconds * 1000;
+    listeners = new ArrayList<>();
+    this.currentMetrics = new AtomicReference<>(new SingleWindowMetrics());
+  }
+
+  @Override
+  public void success() {
+    currentMetrics.get().success();
+    checkTimeout();
```
marsishandsome, line 41: @birdstorm @zz-jason should we use a `Timer` to check timeout?

marsishandsome:
@birdstorm @zz-jason @iosmanthus PTAL, I will add metrics (to Prometheus) in another PR>

src/main/java/org/tikv/common/failsafe/CircuitBreakerImpl.java
```diff
@@ -0,0 +1,187 @@
...
+    };
+  }
+
+  @Override
+  public CircuitBreakerMetrics getMetrics() {
+    return metrics;
+  }
+
+  @Override
+  public boolean allowRequest() {
```
zz-jason, line 99: can we use `isOpen()` to replace this function? for example, change the implementation of `isOpen()` to:

```java
public boolean isOpen() {
    return this.enable & circuitOpened.get() >= 0;
}
```

src/main/java/org/tikv/common/failsafe/CircuitBreakerImpl.java
```diff
@@ -0,0 +1,187 @@
...
+    return !isOpen();
+  }
+
+  @Override
+  public boolean isOpen() {
+    return circuitOpened.get() >= 0;
+  }
+
+  @Override
+  public Status getStatus() {
```
zz-jason, line 112: seems this function is only used by the test, can we remove it from the interface definition and make it visible only to test?

src/main/java/org/tikv/common/failsafe/CircuitBreakerMetricsImpl.java
```diff
@@ -0,0 +1,91 @@
...
+    listeners.add(metricsListener);
+  }
+
+  /** Instead of using SingleWindowMetrics, it is better to use RollingWindowMetrics. */
+  static class SingleWindowMetrics {
+    private final long startMS = System.currentTimeMillis();
+    private final AtomicLong totalCount = new AtomicLong(0);
+    private final AtomicLong errorCount = new AtomicLong(0);
+
+    public void success() {
```
zz-jason, line 74: how about renaming the two functions to:
```java
public void recordSuccess() { .. }
public void recordFailure() { .. }
```

src/main/java/org/tikv/common/failsafe/CircuitBreakerMetricsImpl.java
zz-jason: the current method is a little heavy because the `System.currentTimeMillis()`, would it be better to use `timer`?

src/main/java/org/tikv/common/failsafe/HealthCounts.java
```diff
@@ -0,0 +1,56 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.tikv.common.failsafe;
+
+public class HealthCounts {
+  private final long totalCount;
+  private final long errorCount;
+  private final int errorPercentage;
```
zz-jason, line 21: should it be a double value?

src/main/java/org/tikv/common/failsafe/CircuitBreakerImpl.java
```diff
@@ -0,0 +1,187 @@
...
+          // we are not past the minimum error threshold for the stat window,
+          // so no change to circuit status.
+          // if it was CLOSED, it stays CLOSED
+          // if it was half-open, we need to wait for some successful command executions
+          // if it was open, we need to wait for sleep window to elapse
+        } else {
+          // our failure rate is too high, we need to set the state to OPEN
+          close2Open();
+        }
+      }
```
zz-jason, line 89: why not simplifying it to:
```java
// Open circuit breaker because the failure is too high. Do nothing for other
// conditions, stay in the current state no matter what it is.
if hc.getTotalRequests() >= requestVolumeThreshold && hc.getErrorPercentage() >= errorThresholdPercentage {
    close2Open();
}
```

src/main/java/org/tikv/common/failsafe/CircuitBreakerImpl.java
```diff
@@ -0,0 +1,187 @@
...
+  }
+
+  @Override
+  public void markAttemptFailure() {
+    halfOpen2Open();
+  }
+
+  @Override
+  public boolean attemptExecution() {
+    if (!isOpen()) {
```
zz-jason, line 130: should we use `allowRequest()`?

src/main/java/org/tikv/common/ConfigUtils.java
```diff
@@ -92,6 +92,18 @@
   public static final String TIKV_KEY_CERT_CHAIN = "tikv.key_cert_chain";
   public static final String TIKV_KEY_FILE = "tikv.key_file";
 
+  public static final String TiKV_CIRCUIT_BREAK_ENABLE = "tikv.circuit.break.enable";
```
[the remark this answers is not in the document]
marsishandsome: https://github.com/Netflix/Hystrix/blob/master/hystrix-core/src/main/java/com/netflix/hystrix/HystrixCommandProperties.java#L67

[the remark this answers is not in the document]
marsishandsome: `circuit breaker` is a common usage.

src/main/java/org/tikv/common/failsafe/CircuitBreakerMetricsImpl.java
marsishandsome: ok

src/main/java/org/tikv/common/ConfigUtils.java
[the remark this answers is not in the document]
marsishandsome: rename `tikv.circuit.break` to `tikv.circuit_break`

src/main/java/org/tikv/common/failsafe/CircuitBreakerImpl.java
marsishandsome: `allowRequest` is better. I prefer to delete `isOpen` from the `CircuitBreaker` interface.

src/main/java/org/tikv/common/failsafe/CircuitBreakerImpl.java
marsishandsome: sure

src/main/java/org/tikv/common/failsafe/CircuitBreakerImpl.java
marsishandsome: nice

src/main/java/org/tikv/common/failsafe/HealthCounts.java
marsishandsome: It should be a double value.
In this case, it's enough to use an integer to represent `Percentage`.

src/main/java/org/tikv/common/failsafe/CircuitBreakerImpl.java
marsishandsome: The only reason is that I want to add the comments in a suitable location.

src/main/java/org/tikv/raw/BaseRawKVClient.java
```diff
@@ -0,0 +1,272 @@
...
+
+import com.google.protobuf.ByteString;
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+import org.tikv.common.util.Pair;
+import org.tikv.common.util.ScanOption;
+import org.tikv.kvproto.Kvrpcpb;
+
+public interface BaseRawKVClient extends AutoCloseable {
```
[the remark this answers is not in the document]
marsishandsome: > How about keeping the old `RawKVClient`

It is kept.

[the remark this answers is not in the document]
marsishandsome: move `CircuitBreaker` to package `org.tikv.service.failsafe`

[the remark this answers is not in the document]
marsishandsome: rename `CircuitBreakerRawKVClient` to `SmartRawKVClient` (dualwriter can also added to `SmartRawKVClient` and enabled by config)

src/main/java/org/tikv/service/failsafe/CircuitBreakerMetricsImpl.java
```diff
@@ -0,0 +1,133 @@
...
+  @Override
+  public void recordFailure() {
+    currentMetrics.get().recordFailure();
+  }
+
+  private void checkTimeout() {
+    SingleWindowMetrics singleWindowMetrics = currentMetrics.get();
+    if (System.currentTimeMillis() >= singleWindowMetrics.getStartMS() + windowInMS) {
+      if (currentMetrics.compareAndSet(singleWindowMetrics, new SingleWindowMetrics())) {
+        logger.info("window timeout, reset SingleWindowMetrics");
```
zz-jason, line 72: this log would be printed once a second, it's too frequent.

src/main/java/org/tikv/service/failsafe/CircuitBreakerMetricsImpl.java
```diff
@@ -0,0 +1,133 @@
...
+  }
+
+  @Override
+  public void recordFailure() {
+    currentMetrics.get().recordFailure();
+  }
+
+  private void checkTimeout() {
+    SingleWindowMetrics singleWindowMetrics = currentMetrics.get();
+    if (System.currentTimeMillis() >= singleWindowMetrics.getStartMS() + windowInMS) {
```
zz-jason, line 70: hope this can reduce code indent and improve readabillity.
```java
    if (System.currentTimeMillis() < singleWindowMetrics.getStartMS() + windowInMS) {
      return;
    }
    if (currentMetrics.compareAndSet(singleWindowMetrics, new SingleWindowMetrics())) {
      ...
    }
```

src/main/java/org/tikv/service/failsafe/CircuitBreakerImpl.java
```diff
@@ -0,0 +1,201 @@
...
+    this.requestVolumeThreshold = requestVolumeThreshold;
+    this.sleepWindowInSeconds = sleepWindowInSeconds;
+    this.attemptRequestCount = attemptRequestCount;
+    this.metrics = new CircuitBreakerMetricsImpl(windowInSeconds);
+    this.metrics.addListener(getMetricsListener());
+  }
+
+  private MetricsListener getMetricsListener() {
+    return hc -> {
+      logger.info("onNext " + hc.toString());
```
zz-jason, line 79: ditto, it may periodically print useless information. I think we only need to log a warning message when the circuit breaker is open, half-open, or closed. Other messages can be debug-level.

src/main/java/org/tikv/service/failsafe/CircuitBreaker.java
```diff
@@ -0,0 +1,60 @@
...
+  boolean attemptExecution();
+
+  /** Invoked on successful executions as part of feedback mechanism when in a half-open state. */
+  void recordAttemptSuccess();
+
+  /** Invoked on unsuccessful executions as part of feedback mechanism when in a half-open state. */
+  void recordAttemptFailure();
+
+  /** Get the Circuit Breaker Metrics Object. */
+  CircuitBreakerMetrics getMetrics();
```
zz-jason, line 59: it's a little complicated that the users need to care about how to use circuit metrics and breaker, can we only expose the breaker to the outside?

src/main/java/org/tikv/service/failsafe/CircuitBreakerMetricsImpl.java
```diff
@@ -0,0 +1,115 @@
...
+  private final ScheduledExecutorService scheduler;
+  private static final int SCHEDULER_INITIAL_DELAY = 1000;
+  private static final int SCHEDULER_PERIOD = 1000;
+
+  public CircuitBreakerMetricsImpl(int windowInSeconds) {
+    this.windowInMS = windowInSeconds * 1000;
+    this.listeners = new ArrayList<>();
+    this.currentMetrics = new AtomicReference<>(new SingleWindowMetrics());
+
+    scheduler =
```
zz-jason, line 46: should this background task be managed by the circuit breaker instead of the circuit metrics?

src/main/java/org/tikv/raw/SmartRawKVClient.java
```diff
@@ -0,0 +1,291 @@
...
+  public void deleteRange(ByteString startKey, ByteString endKey) {
+    callWithCircuitBreaker("deleteRange", () -> client.deleteRange(startKey, endKey));
+  }
+
+  @Override
+  public void deletePrefix(ByteString key) {
+    callWithCircuitBreaker("deletePrefix", () -> client.deletePrefix(key));
+  }
+
+  <T> T callWithCircuitBreaker(String funcName, Function1<T> func) {
```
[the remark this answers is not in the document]
marsishandsome: then it will become nested `try catch` with normal logic.

src/main/java/org/tikv/raw/SmartRawKVClient.java
```diff
@@ -0,0 +1,291 @@
...
+        circuitBreakerMetrics.recordFailure();
+        throw e;
+      }
+    } else if (circuitBreaker.attemptExecution()) {
+      logger.info("attemptExecution");
+      try {
+        T result = func.apply();
+        circuitBreakerMetrics.recordSuccess();
+        circuitBreaker.recordAttemptSuccess();
+        logger.info("markSuccess");
```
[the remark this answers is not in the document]
marsishandsome: trouble shooting

src/main/java/org/tikv/service/failsafe/CircuitBreakerMetricsImpl.java
```diff
@@ -0,0 +1,115 @@
...
+  public void recordSuccess() {
+    currentMetrics.get().recordSuccess();
+  }
+
+  @Override
+  public void recordFailure() {
+    currentMetrics.get().recordFailure();
+  }
+
+  private void checkTimeout() {
```
zz-jason, line 68: how about renaming it to `onReachCircuitWindow()`?

src/main/java/org/tikv/raw/SmartRawKVClient.java
```diff
@@ -0,0 +1,291 @@
...
+        return result;
+      } catch (Exception e) {
+        circuitBreakerMetrics.recordFailure();
+        circuitBreaker.recordAttemptFailure();
+        logger.info("markNonSuccess");
+        throw e;
+      }
+    } else {
+      logger.warn("Circuit Breaker Opened");
```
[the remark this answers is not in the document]
marsishandsome: if Circuit Breaker happens, the whole cluster may be down, so the `numerous warning` is not a problem.

src/main/java/org/tikv/service/failsafe/HealthCounts.java
```diff
@@ -0,0 +1,56 @@
...
+  @Override
+  public String toString() {
+    return "HealthCounts{"
+        + "totalCount="
+        + totalCount
+        + ", errorCount="
+        + errorCount
+        + ", errorPercentage="
+        + errorPercentage
+        + '}';
```
zz-jason, line 54: how about reformating it to: `traffic stats in d% seconds: total %d, error %d, error rate: %f`, it's shorter and contains more info than before.

src/main/java/org/tikv/raw/SmartRawKVClient.java
```diff
@@ -0,0 +1,291 @@
...
+  private static final Counter CIRCUIT_BREAKER_OPENED =
+      Counter.build()
+          .name("client_java_smart_raw_circuit_breaker_opened")
+          .help("client smart raw circuit breaker opened.")
+          .labelNames("type")
+          .register();
+
+  private final RawKVClientBase client;
+  private final CircuitBreaker circuitBreaker;
+  private final CircuitBreakerMetrics circuitBreakerMetrics;
```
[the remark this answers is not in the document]
marsishandsome: > The circuit breaker is designed to be thread-local.
No.

> every session open or close the circuit breaker individually, am I correct?
Every `TiSession` open or close the circuit breaker individually.

src/main/java/org/tikv/service/failsafe/HealthCounts.java
```diff
@@ -0,0 +1,56 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.tikv.service.failsafe;
+
+public class HealthCounts {
```
zz-jason, line 18: how about renaming it to `TrafficStats`?

src/main/java/org/tikv/raw/SmartRawKVClient.java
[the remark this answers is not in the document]
marsishandsome: > could you add some comments here to explain how circuit breaker and circuit metrics work together to provide the circuit breaking management?

`Client` asks `CircuitBreaker` if requests are allow.
`Client` sends result feedback to `CircuitBreakerMetrics`.
`CircuitBreakerMetrics` tells `CircuitBreaker` the current error rate.

src/main/java/org/tikv/service/failsafe/CircuitBreakerMetricsImpl.java
marsishandsome: `info` level is used. Users can set `warn` or `error` level in prod env.

src/main/java/org/tikv/service/failsafe/CircuitBreakerImpl.java
```diff
@@ -0,0 +1,201 @@
...
+  private static final Counter CIRCUIT_BREAKER_ATTEMPT_COUNTER =
+      Counter.build()
+          .name("client_java_circuit_breaker_attempt_counter")
+          .help("client circuit breaker attempt counter.")
+          .labelNames("type")
+          .register();
+
+  private final boolean enable;
+  private final int windowInSeconds;
+  private final int errorThresholdPercentage;
```
zz-jason, line 38: why not use double to store the error rate threshold, since it would be more concise?

src/main/java/org/tikv/service/failsafe/CircuitBreakerImpl.java
marsishandsome: `info` level is used here. Users can set `warn` or `error` level in prod env.

zz-jason: info level is quite normal, we should carefully print info logs as well.

zz-jason: if it causes too much noise, then it should not be printed in info, or it should be printed in debug level.

src/main/java/org/tikv/service/failsafe/CircuitBreakerImpl.java
marsishandsome: In this case, it's enough to use an integer to represent Percentage.

src/main/java/org/tikv/service/failsafe/HealthCounts.java
marsishandsome: sounds bad.

src/main/java/org/tikv/service/failsafe/HealthCounts.java
marsishandsome: Using IDE auto generated t`oString` implemation makes users easier to understand.

src/main/java/org/tikv/raw/SmartRawKVClient.java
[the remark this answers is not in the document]
zz-jason: could the client only communicate with the circuit breaker? for example:

1. client asks circuit breaker whether the requests are allowed to be executed
2. client sends request success or failure info as feedbacks to circuit breaker

In this way, we can limit the complexity and potential bugs to the circuit breaker itself.

src/main/java/org/tikv/service/failsafe/HealthCounts.java
marsishandsome: `HealthCounts` refes to the implementation of `Hystrix`.

src/main/java/org/tikv/service/failsafe/CircuitBreakerMetricsImpl.java
marsishandsome: why?

src/main/java/org/tikv/service/failsafe/CircuitBreaker.java
marsishandsome: It's easy for us to implement `RollingWindowMetrics` if we separate metrics from breaker.
Another reason is that it's easy to test.

marsishandsome: We should not put every thing into one Class.

src/main/java/org/tikv/raw/SmartRawKVClient.java
```diff
@@ -0,0 +1,291 @@
...
+    try {
+      client.close();
+    } catch (Exception e) {
+      err = e;
+    }
+
+    try {
+      circuitBreaker.close();
+    } catch (Exception e) {
+      err = e;
```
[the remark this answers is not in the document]
marsishandsome: fixed

src/main/java/org/tikv/raw/SmartRawKVClient.java
```diff
@@ -0,0 +1,291 @@
...
+    }
+  }
+
+  public interface Function1<T> {
+    T apply();
+  }
+
+  public interface Function0 {
+    void apply();
+  }
```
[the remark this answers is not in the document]
marsishandsome: it's the best solution i can find.

src/main/java/org/tikv/raw/SmartRawKVClient.java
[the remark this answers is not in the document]
marsishandsome: It's easy for us to implement `RollingWindowMetrics` if we separate metrics from breaker.
Another reason is that it's easy to test.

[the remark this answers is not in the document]
marsishandsome: Put everything into one class is not a good design.

src/main/java/org/tikv/raw/SmartRawKVClient.java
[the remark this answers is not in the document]
marsishandsome: change to debug level

src/main/java/org/tikv/service/failsafe/CircuitBreakerMetricsImpl.java
marsishandsome: change to debug level

src/main/java/org/tikv/service/failsafe/CircuitBreakerImpl.java
marsishandsome: change to debug level

src/main/java/org/tikv/raw/SmartRawKVClient.java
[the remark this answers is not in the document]
marsishandsome: change to debug level

src/main/java/org/tikv/raw/SmartRawKVClient.java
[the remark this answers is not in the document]
marsishandsome: I remove `private final CircuitBreakerMetrics circuitBreakerMetrics;` from `SmartRawKVClient`.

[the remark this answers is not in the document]
marsishandsome: https://github.com/tikv/client-java/pull/358/commits/cddd86fb72aa70390131c6148e742974c54bc6d6

src/main/java/org/tikv/service/failsafe/CircuitBreakerMetricsImpl.java
zz-jason: BTW, we should not start this background thread when the circuit breaker is configured to be disabled.

src/main/java/org/tikv/service/failsafe/CircuitBreakerImpl.java
```diff
@@ -0,0 +1,201 @@
...
+  private final int requestVolumeThreshold;
+  private final int sleepWindowInSeconds;
+  private final int attemptRequestCount;
+
+  private final AtomicLong circuitOpened = new AtomicLong(-1);
+  private final AtomicReference<Status> status = new AtomicReference<>(Status.CLOSED);
+  private final AtomicLong attemptCount = new AtomicLong(0);
+  private final AtomicLong attemptSuccessCount = new AtomicLong(0);
```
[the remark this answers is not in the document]
zz-jason: There are two-time windows and their corresponding traffic stats:

1. the first time window is to record normal traffic stats when the circuit breaker is closed or disabled in the background and triggers circuit breaker, which is limited to `TiKV_CIRCUIT_BREAK_AVAILABILITY_WINDOW_IN_SECONDS` and is called **windowOfClosed**.
2. the second time window to record the attempt traffic stats when the circuit breaker is half-opened, which is called **windowOfHalfOpened**.

these two windows can be posited like this: `windowOfClosed -> (circuit breaker opened) -> (after sleep, circuit breaker half-opened) -> windowOfHalfOpened`

So can we:

1. Define two traffic metrics for these two time windows like the following and remove the atomic variables for attempts:
```java
private final CircuitBreakerMetrics metricsOfClosed;
private final CircuitBreakerMetrics metricsOfHalfOpened
```

2. Make `CircuitBreakerImpl` be the owner of these two windows, don't start the background task to collect `metricsOfClosed` when the circuit breaker is disabled.

src/main/java/org/tikv/service/failsafe/CircuitBreakerImpl.java
```diff
@@ -0,0 +1,201 @@
...
+
+  @Override
+  public boolean allowRequest() {
+    if (!enable) {
+      return true;
+    }
+    return !isOpen();
+  }
+
+  boolean isOpen() {
```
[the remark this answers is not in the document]
marsishandsome: it is used in test.

src/main/java/org/tikv/service/failsafe/CircuitBreakerImpl.java
```diff
@@ -0,0 +1,201 @@
...
+    return circuitOpened.get() >= 0;
+  }
+
+  Status getStatus() {
```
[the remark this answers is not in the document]
marsishandsome: it is used in test.

src/main/java/org/tikv/service/failsafe/CircuitBreakerImpl.java
marsishandsome: `Hystrix` use `errorThresholdPercentage` as integer type [0, 100].

src/main/java/org/tikv/raw/SmartRawKVClient.java
```diff
@@ -0,0 +1,274 @@
...
+      try {
+        T result = func.apply();
+        circuitBreaker.getMetrics().recordSuccess();
+        return result;
+      } catch (Exception e) {
+        circuitBreaker.getMetrics().recordFailure();
+        throw e;
+      }
+    } else if (circuitBreaker.attemptExecution()) {
+      logger.debug("attemptExecution");
```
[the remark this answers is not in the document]
marsishandsome: log4j will put the class name as prefix.

src/main/java/org/tikv/service/failsafe/CircuitBreakerImpl.java
```diff
@@ -0,0 +1,201 @@
...
+
+  private static final Counter CIRCUIT_BREAKER_ATTEMPT_COUNTER =
+      Counter.build()
+          .name("client_java_circuit_breaker_attempt_counter")
+          .help("client circuit breaker attempt counter.")
+          .labelNames("type")
+          .register();
+
+  private final boolean enable;
+  private final int windowInSeconds;
```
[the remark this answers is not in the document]
marsishandsome: yes, but it is better to leave it here.

src/main/java/org/tikv/service/failsafe/CircuitBreakerImpl.java
```diff
@@ -0,0 +1,201 @@
...
+  }
+
+  private boolean isAfterSleepWindow() {
+    final long circuitOpenTime = circuitOpened.get();
+    final long currentTime = System.currentTimeMillis();
+    final long sleepWindowTime = (long) sleepWindowInSeconds * 1000;
+    return currentTime >= circuitOpenTime + sleepWindowTime;
+  }
+
+  private void close2Open() {
```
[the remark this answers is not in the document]
marsishandsome: why does the function order matter with readability?

src/main/java/org/tikv/service/failsafe/CircuitBreakerImpl.java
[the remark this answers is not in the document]
marsishandsome: > do we really need to set these variables as atomic variables? 

yes, in order to make `CircuitBreakerImpl` thread-safe.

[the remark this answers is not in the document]
marsishandsome: > we add the 'final' keyword to these fields but their values are changed during the execution time, seems a little counterintuitive.

yes, we should add `final` and the object reference does not change. The object content does change.

src/main/java/org/tikv/service/failsafe/CircuitBreakerImpl.java
```diff
@@ -0,0 +1,201 @@
...
+    this(
+        conf.isCircuitBreakEnable(),
+        conf.getCircuitBreakAvailabilityWindowInSeconds(),
+        conf.getCircuitBreakAvailabilityErrorThresholdPercentage(),
+        conf.getCircuitBreakAvailabilityRequestVolumnThreshold(),
+        conf.getCircuitBreakSleepWindowInSeconds(),
+        conf.getCircuitBreakAttemptRequestCount());
+  }
+
+  public CircuitBreakerImpl(
```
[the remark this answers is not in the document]
marsishandsome: it is designed for test.

src/main/java/org/tikv/service/failsafe/CircuitBreakerImpl.java
```diff
@@ -0,0 +1,201 @@
...
+        // if some of the executing commands fail, the status will transition to OPEN
+        open2HalfOpen();
+        return attemptCount.incrementAndGet() <= attemptRequestCount;
+      } else {
+        return false;
+      }
+    }
+  }
+
+  private boolean isAfterSleepWindow() {
```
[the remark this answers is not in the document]
marsishandsome: `isAfterSleepWindow` is only be called from `attemptExecution`, which means the `CircuitBreaker` is opened.

In this case, the performance is acceptable.

src/main/java/org/tikv/service/failsafe/CircuitBreakerMetricsImpl.java
marsishandsome: fixed in https://github.com/tikv/client-java/pull/358/commits/5fcf19b7c70f938f938742531498df6f2f1b454a

birdstorm (approved):
LGTM

zz-jason (approved):
LGTM

ti-srebot (approved):

PR #359, opened 2021-12-03
Fix NullPointerException in getStore

birdstorm:
If `callWIthRetry` returns a null response, `getStore` will throw NullPointerException.

### DIFF
--- a/src/main/java/org/tikv/common/PDClient.java
+++ b/src/main/java/org/tikv/common/PDClient.java
@@ -356,9 +356,16 @@ public class PDClient extends AbstractGRPCClient<PDBlockingStub, PDFutureStub>
 
   @Override
   public Store getStore(BackOffer backOffer, long storeId) {
-    return callWithRetry(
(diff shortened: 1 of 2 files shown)


src/main/java/org/tikv/common/PDClient.java
```diff
@@ -356,9 +356,16 @@ private boolean isScatterRegionFinish(GetOperatorResponse resp) {
...
+    GetStoreResponse resp =
+        callWithRetry(
+            backOffer,
+            PDGrpc.getGetStoreMethod(),
+            buildGetStoreReq(storeId),
+            buildPDErrorHandler());
+    if (resp != null) {
+      return resp.getStore();
+    }
+    return null;
```
zz-jason, line 368: could you add some comments about:

1. the possible scenarios where get store fails with a null response
2. why not throw an exception in those scenarios

Little-Wallace (approved):
LGTM

marsishandsome (approved):
LGTM

ti-srebot (approved):

PR #366, opened 2021-12-07
fix seekLeaderStore NPE

iosmanthus:

This pull request fixes the NPE triggered by the double retry of `seekLeaderStore`. These buggy lines would also introduce another issue like `peer id not match` because the meta of the region is not updated.

Once we change detected a reachable store we need to update the peer information of the region and retry in the most outside code.

### Check List 

Tests 

 - Integration test

Code changes

 - Has persistent data change

Side effects

 - Increased code complexity

Related changes


marsishandsome (approved):
LGTM

zz-jason:
@iosmanthus please remove the necessary checks in PR description

zz-jason (changes requested):
could you add some tests?

iosmanthus:
> could you add some tests?

It is really hard to inject the error to trigger a double retry of seekLeaderStore. We can only use integration to cover this case.

zz-jason (approved):
LGTM

ti-srebot (approved):

PR #368, opened 2021-12-08
warm up RawKVClient while creating it

iosmanthus:

We find some initialization work that needs to do in the first gRPC call in `RawKVClient` which might result in a timeout in the first few gRPC calls.

First run:

image

Second run:

image

To solve this problem, we may launch a `raw_get` request to warm up the runtime.

### Check List 

Tests 

 - Integration test
 - No code

Code changes

 - Has persistent data change

Side effects

 - Increased code complexity

Related changes


birdstorm:
IMHO not suitable to warm up the client by calling `get()` when initializing.

Issue #375, opened 2021-12-09
the first requests is very slow in RawKVClient


zz-jason:
**Is your feature request related to a problem? Please describe.**

The first request 

```json
{
    "start":"17:45:35.667",
    "end":"17:45:35.720",
    "duration":"53ms",
    "func":"get",
    "region":"{Region[8590814505] ConfVer[7164] Version[165] Store[8589948024] KeyRange[xxxxx]:[xxxxx]}",
    "key":"xxxxx",
    "spans":[
        {
            "name":"getRegionByKey",
            "start":"17:45:35.713",
            "end":"17:45:35.715",
            "duration":"2ms"
        },
        {
            "name":"callWithRetry pdpb.PD/GetRegion",
            "start":"17:45:35.713",
            "end":"17:45:35.715",
            "duration":"2ms"
        },
        {
            "name":"gRPC pdpb.PD/GetRegion",
            "start":"17:45:35.713",
            "end":"17:45:35.715",
            "duration":"2ms"
        },
        {
            "name":"callWithRetry tikvpb.Tikv/RawGet",
            "start":"17:45:35.716",
            "end":"17:45:35.720",
            "duration":"4ms"
        },
        {
            "name":"gRPC tikvpb.Tikv/RawGet",
            "start":"17:45:35.716",
            "end":"17:45:35.720",
            "duration":"4ms"
        }
    ]
}
```
**Describe the solution you'd like**

make them fast as the client is running for a long time

**Describe alternatives you've considered**

N/A

**Additional context**

N/A

PR #383, opened 2021-12-09
[to #375] use scanRegions request to warm up client

birdstorm:
use scanRegions request with limit to warm up client.

new configurations:
```
tikv.warm_up.enable default=true
tikv.grpc.warm_up_timeout_in_ms default=5000ms
```


src/main/java/org/tikv/common/TiSession.java
```diff
@@ -117,11 +118,21 @@ private synchronized void warmUp() {
...
+                ByteString.EMPTY,
+                conf.getScanRegionsLimit());
+        for (Pdpb.Region region : regions) {
+          regionManager.insertRegionToCache(
+              regionManager.createRegion(region.getRegion(), ConcreteBackOffer.newGetBackOff()));
+        }
+        startKey = regions.get(regions.size() - 1).getRegion().getEndKey();
       } while (!startKey.isEmpty());
 
       RawKVClient rawKVClient = createRawClient();
```
zz-jason, line 138: should we close the client after the warmup is finished?

zz-jason (commented):
could you add some test for this PR?

src/main/java/org/tikv/common/TiSession.java
birdstorm: done

src/test/java/org/tikv/common/TiSessionTest.java
```diff
@@ -122,4 +125,29 @@ private void doCloseTest(boolean now, long timeoutMS) throws Exception {
...
+    TiConfiguration conf = createTiConfiguration();
+    conf.setWarmUpEnable(true);
+    doTest(conf);
+    conf.setWarmUpEnable(false);
+    doTest(conf);
+  }
+
+  private void doTest(TiConfiguration conf) throws Exception {
+    session = TiSession.create(conf);
+    long start = System.currentTimeMillis();
```
zz-jason, line 140: is it better to use System.nanoTime()?

src/test/java/org/tikv/common/TiSessionTest.java
```diff
@@ -122,4 +125,29 @@ private void doCloseTest(boolean now, long timeoutMS) throws Exception {
...
     }
   }
+
+  @Test
+  public void warmUpTest() throws Exception {
+    TiConfiguration conf = createTiConfiguration();
+    conf.setWarmUpEnable(true);
+    doTest(conf);
+    conf.setWarmUpEnable(false);
+    doTest(conf);
```
zz-jason, line 135: how about:

```java
elapsedWithWarmupMS  = doTest(conf)
conf.setWarmUpEnable(false);
elapsedWithoutWarmupMS = doTest(conf);
assertTrue(elapsedWithWarmupMS < elapsedWithoutWarmupMS)
```

src/main/java/org/tikv/common/PDClient.java
```diff
@@ -341,6 +341,27 @@ private boolean isScatterRegionFinish(GetOperatorResponse resp) {
...
   }
 
+  @Override
+  public List<Pdpb.Region> scanRegions(
+      BackOffer backOffer, ByteString startKey, ByteString endKey, int limit) {
+    // no need to backoff because ScanRegions is just for optimization
+    // introduce a warm-up timeout for ScanRegions requests
+    PDGrpc.PDBlockingStub stub =
+        getBlockingStub().withDeadlineAfter(conf.getWarmUpTimeout(), TimeUnit.MILLISECONDS);
+    Pdpb.ScanRegionsRequest request =
```
marsishandsome, line 351: should we use `callWithRetry` here?

birdstorm: I don't think we should use callWithRetry here because we don't need to retry when the request fails.

src/test/java/org/tikv/common/TiSessionTest.java
birdstorm: I think it is not necessary in unit test.

src/main/java/org/tikv/common/PDClient.java
marsishandsome: then i suggest to rename `scanRegions` to `scanRegionsWithoutRetry`.
The gRPC methods do retry as default.

marsishandsome (approved):
LGTM

src/test/java/org/tikv/common/TiSessionTest.java
zz-jason: @birdstorm PTAL at this comment

birdstorm: @zz-jason PTAL

zz-jason (approved):
LGTM

ti-srebot (approved):

PR #408, opened 2021-12-15
[close #397] add git commit hash and version info while TiSession create

### LINKED ISSUE #397: print version number, commit hash, and config items after java client initialized
**Is your feature request related to a problem? Please describe.**

when diagnosis performance or other client-related issues, we may need to collect the following information to locate the root cause:

- welcome message, helps us to know when the client was rebooted.
- the version number
- commit hash, double-check the version
- config items, what config the users are using, especially the config values that is different from the default one.

**Describe the solution you'd like**

When the client is initialized, print a **INFO** log like the following as the first log:

```
Welcome to Java Client! Version: ${version number}@${commit hash}, Config: ${config items in JSON}
```

**Describe alternatives you've considered**

It would be better if we can also print this information via the jar package

**Additional context**

N/A

iosmanthus:

Add info logs while `Tisession` create

Issue Number: close #397

Problem Description: **TBD**

### Check List for Tests

### Related changes


src/main/java/org/tikv/common/TiSession.java
```diff
@@ -103,6 +103,18 @@ public TiSession(TiConfiguration conf) {
     warmUp();
     this.circuitBreaker = new CircuitBreakerImpl(conf);
     logger.info("TiSession initialized in " + conf.getKvMode() + " mode");
+    printVersionInfo();
+  }
+
+  private void printVersionInfo() {
```
zz-jason, line 109: how about refactoring this function to: 
```java
private void getVersionInfo() String {
}
```

then log the info in `TiSession` and add a UT for this function.

pom.xml
```diff
@@ -551,6 +551,30 @@
                     <autoReleaseAfterClose>false</autoReleaseAfterClose>
                 </configuration>
             </plugin>
+            <plugin>
+                <groupId>io.github.git-commit-id</groupId>
+                <artifactId>git-commit-id-maven-plugin</artifactId>
+                <version>4.9.9</version>
+                <executions>
+                    <execution>
+                        <id>get-the-git-infos</id>
```
birdstorm, line 560: I would like to ask if this execution still works when being used as a maven dependency?

src/main/java/org/tikv/common/TiSession.java
iosmanthus: It's hard to write suck UT to assert the commit hash.

iosmanthus: Now, I capture these logic into `VersionInfo`, PTAL.

src/main/java/org/tikv/common/TiSession.java
```diff
@@ -103,6 +120,22 @@ public TiSession(TiConfiguration conf) {
...
+  private VersionInfo getVersionInfo() {
+    VersionInfo info;
+    try {
+      final Properties properties = new Properties();
+      properties.load(this.getClass().getClassLoader().getResourceAsStream("git.properties"));
+      String version = properties.getProperty("git.build.version");
+      String commitHash = properties.getProperty("git.commit.id.full");
+      info = new VersionInfo(version, commitHash);
+    } catch (Exception e) {
+      logger.info("Fail to read package info: " + e.getMessage());
```
marsishandsome, line 135:
```suggestion
      logger.error("Fail to read package info: " + e.getMessage());
```

iosmanthus: conflicted with: https://github.com/tikv/client-java/pull/408#discussion_r770188489. Personally speaking, I think the error log level could help us discover some building issues.

src/main/java/org/tikv/common/TiSession.java
```diff
@@ -103,6 +120,22 @@ public TiSession(TiConfiguration conf) {
     warmUp();
     this.circuitBreaker = new CircuitBreakerImpl(conf);
     logger.info("TiSession initialized in " + conf.getKvMode() + " mode");
+    logger.info("Welcome to TiKV Java Client {}", getVersionInfo());
```
zz-jason, line 123: It would print a welcome message when every new `TiSession` is created. Considering that an application can use multiple `TiSession`s, is there a better place to print this message?

pom.xml
iosmanthus: Yes, the git commit hash is get in compile phase.

src/main/java/org/tikv/common/TiSession.java
iosmanthus: Done.

zz-jason (approved):
LGTM

ti-srebot (approved):

Issue #404, opened 2021-12-14
[BUG] SlowLog endTime may < startTime


marsishandsome:
https://stackoverflow.com/questions/2978598/will-system-currenttimemillis-always-return-a-value-previous-calls

![image](https://user-images.githubusercontent.com/5574887/145984856-812c4fd4-4a29-4ce3-9dec-99491b7fa194.png)

PR #434, opened 2021-12-24
[close #433] fix calling getStoreById without backoffer

### LINKED ISSUE #433: Timeout in seekLeaderStore is not limited
Here is a SlogLog generated by the case `kill az with pd leader`.

```
{"start":"21:12:36.806","end":"21:12:42.299","duration":"5493ms","error":"retry is exhausted.","func":"get","region":"{Region[12358] ConfVer[547] Version[20] Store[20] KeyRange[indexInfo_:_pf01_:_APD01_:_3500770017512150000]:[indexInfo_:_pf01_:_APD01_:_3501970010589630000]}","key":"indexInfo_:_pf01_:_APD01_:_3501910019979570000","spans":[{"name":"getRegionByKey","start":"21:12:36.806","end":"21:12:36.806","duration":"0ms"},{"name":"callWithRetry tikvpb.Tikv/RawGet","start":"21:12:36.806","end":"21:12:42.298","duration":"5492ms"},{"name":"gRPC tikvpb.Tikv/RawGet","start":"21:12:36.806","end":"21:12:36.956","duration":"150ms"},{"name":"backoff BoCheckTimeout","start":"21:12:36.956","end":"21:12:36.956","duration":"0ms"},{"name":"backoff BoCheckTimeout","start":"21:12:36.956","end":"21:12:36.956","duration":"0ms"},{"name":"seekLeaderStore","start":"21:12:36.956","end":"21:12:42.298","duration":"5342ms"},{"name":"backoff BoRegionMiss","start":"21:12:42.298","end":"N/A","duration":"N/A"}]}
```

Timeout in seekLeaderStore is not limited
```
{
            "name":"seekLeaderStore",
            "start":"21:12:36.956",
            "end":"21:12:42.298",
            "duration":"5342ms"
        }
```

marsishandsome:
close #433

Timeout in seekLeaderStore is not limited.

We should use `backoffer` for all the gRPC calls.

### DIFF
--- a/src/main/java/org/tikv/common/TiSession.java
+++ b/src/main/java/org/tikv/common/TiSession.java
@@ -161,16 +161,16 @@ public class TiSession implements AutoCloseable {
   }
 
   private synchronized void warmUp() {
-    long warmUpStartTime = System.currentTimeMillis();
+    long warmUpStartTime = System.nanoTime();
+    BackOffer backOffer = ConcreteBackOffer.newRawKVBackOff();
     try {
       this.client = getPDClient();
       this.regionManager = getRegionManager();
-      List<Metapb.Store> stores = this.client.getAllStores(ConcreteBackOffer.newGetBackOff());
+      List<Metapb.Store> stores = this.client.getAllStores(backOffer);
       // warm up store cache
       for (Metapb.Store store : stores) {
         this.regionManager.updateStore(
-            null,
-            new TiStore(this.client.getStore(ConcreteBackOffer.newGetBackOff(), store.getId())));
+            null, new TiStore(this.client.getStore(backOffer, store.getId())));
       }
 
       // use scan region to load region cache with limit
@@ -178,17 +178,14 @@ public class TiSession implements AutoCloseable {
       do {
         List<Pdpb.Region> regions =
             regionManager.scanRegions(
-                ConcreteBackOffer.newGetBackOff(),
-                startKey,
-                ByteString.EMPTY,
-                conf.getScanRegionsLimit());
+                backOffer, startKey, ByteString.EMPTY, conf.getScanRegionsLimit());
         if (regions == null || regions.isEmpty()) {
           // something went wrong, but the warm-up process could continue
           break;
         }
         for (Pdpb.Region region : regions) {
           regionManager.insertRegionToCache(
-              regionManager.createRegion(region.getRegion(), ConcreteBackOffer.newGetBackOff()));
+              regionManager.createRegion(region.getRegion(), backOffer));
         }
         startKey = regions.get(regions.size() - 1).getRegion().getEndKey();
       } while (!startKey.isEmpty());
@@ -211,7 +208,8 @@ public class TiSession implements AutoCloseable {
       logger.info("warm up fails, ignored ", e);
     } finally {
       logger.info(
-          String.format("warm up duration %d ms", System.currentTimeMillis() - warmUpStartTime));
+          String.format(
+              "warm up duration %d ms", (System.nanoTime() - warmUpStartTime) / 1_000_000));
     }
   }
 
--- a/src/main/java/org/tikv/common/util/ConcreteBackOffer.java
+++ b/src/main/java/org/tikv/common/util/ConcreteBackOffer.java
@@ -168,8 +168,6 @@ public class ConcreteBackOffer implements BackOffer {
   }
 
   public boolean canRetryAfterSleep(BackOffFunction.BackOffFuncType funcType, long maxSleepMs) {
-    SlowLogSpan slowLogSpan = getSlowLog().start("backoff " + funcType.name());
-    Histogram.Timer backOffTimer = BACKOFF_DURATION.labels(funcType.name()).startTimer();
     BackOffFunction backOffFunction =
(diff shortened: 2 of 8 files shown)


birdstorm (approved):
LGTM

src/main/java/org/tikv/common/TiSession.java
```diff
@@ -159,23 +159,46 @@ private static VersionInfo getVersionInfo() {
 
   private synchronized void warmUp() {
     long warmUpStartTime = System.currentTimeMillis();
+    warmUpStore();
+    warmUpRegion();
+    warmUpGrpc();
+    logger.info(
+        String.format("warm up duration %d ms", System.currentTimeMillis() - warmUpStartTime));
```
zz-jason, line 166: should we use `System.nanoTime()` to avoid issues #404 happen again?

src/main/java/org/tikv/common/TiSession.java
```diff
@@ -159,23 +159,46 @@ private static VersionInfo getVersionInfo() {
...
       for (Metapb.Store store : stores) {
         this.regionManager.updateStore(
-            null,
-            new TiStore(this.client.getStore(ConcreteBackOffer.newGetBackOff(), store.getId())));
+            null, new TiStore(this.client.getStore(backOffer, store.getId())));
       }
-      ByteString startKey = ByteString.EMPTY;
+    } catch (Exception e) {
+      // ignore error
+      logger.info("warm up store fails, ignored ", e);
```
zz-jason, line 182: since the log is printed, it's better to print the failure reason from the exception to help to locate potential issues. warmup failures can be ignored by users, but cannot be ignored by us.

src/main/java/org/tikv/common/TiSession.java
```diff
@@ -159,23 +159,46 @@ private static VersionInfo getVersionInfo() {
 
   private synchronized void warmUp() {
     long warmUpStartTime = System.currentTimeMillis();
+    warmUpStore();
```
zz-jason, line 162: would it be better to handle exceptions in this function since all the exceptions are ignored? for example:

```java
try {
  warmUpStore();
  warmUpRegion();
  ...
} catch () {
  ...
}
```

src/main/java/org/tikv/common/util/ConcreteBackOffer.java
```diff
@@ -185,6 +184,7 @@ public boolean canRetryAfterSleep(BackOffFunction.BackOffFuncType funcType, long
       }
     }
 
+    SlowLogSpan slowLogSpan = getSlowLog().start("backoff " + funcType.name());
```
zz-jason, line 187: why move it here? should `BACKOFF_DURATION` also be moved?

zz-jason (commented):
please add some tests.

src/main/java/org/tikv/common/TiSession.java
marsishandsome: we do not need an accurate measure here.

src/main/java/org/tikv/common/TiSession.java
marsishandsome: the current code will print the `e` in log file.

src/main/java/org/tikv/common/TiSession.java
marsishandsome: if `warmUpStore` failes, `warmUpRegion` will still run.

src/main/java/org/tikv/common/util/ConcreteBackOffer.java
marsishandsome: there is a `return` in line 184, which will cause ` "end":"N/A",`.
it's hard to understand it.

src/main/java/org/tikv/common/TiSession.java
zz-jason: since you are logging the time, and System.currentTimeMillis() can be misleading, I think we should avoid using it to record time durations.

src/main/java/org/tikv/common/util/ConcreteBackOffer.java
zz-jason: got. then why not move `BACKOFF_DURATION` to here as well?

src/main/java/org/tikv/common/TiSession.java
zz-jason: ok

src/main/java/org/tikv/common/util/ConcreteBackOffer.java
marsishandsome: good idea

zz-jason (approved):
LGTM

ti-srebot (approved):

src/main/java/org/tikv/common/util/ConcreteBackOffer.java
iosmanthus: Moving these lines to here might cause we can't see the backoff in slow log.

iosmanthus: We may end this span while returning though we will see this backoff end shortly but this span is still there.

iosmanthus:
Rest LGTM

Issue #477, opened 2022-01-11
introduce tracing for client-java


iosmanthus:
## Feature Request

**Is your feature request related to a problem? Please describe:**

For now, it is hard to debug a request from clients to TiKV servers. We could only use region/store ids and log timestamps to trace the life of a request which is inefficient. It is hard to tell if two logs with the same region/store ids info are produced by the same request because clients may retry a gRPC request a few times. We need a mechanism to track the lifetime of a request.

**Describe the feature you'd like:**

Introduce tracing for client-java and TiKV. Currently, TiKV has a pull request for the basic infrastructure to this feature: https://github.com/tikv/tikv/pull/11824.

From the client perspective, adding a tracing context to our request should be enough to track a request:
https://github.com/pingcap/kvproto/blob/3fa8fa04f898c8b2f5f9b7b8576f38016cb2d5e6/proto/trace.proto#L14-L19
```protobuf
message TraceContext {
    // The id that is able to identify a unique request. It's usually a UUID.
    uint64 trace_id = 1;
    // The span that represents the caller's calling procedural.
    uint64 parent_id = 2;
}
```

The tracing could be introduced in two phases:

1. **tracing by logs**
While constructing a request in client-java, we should attach a generated UUID and `parent_id = 0` (indicating it is the root caller of a request) to the tikv gRPC context. In addition, `trace_id` should be attached to the slow log of the request(both of client and server) as a property like a region id and a store id. While debugging requests, we could grep all the logs of clients and servers with the `trace_id` as the primary key. This process might be a little bit annoying, but it works for most scenarios.

2. **tracing by jaeger or something else**
Instrument the client-java with the standard of OpenTracing, and collect the span with components like [jaeger](https://www.jaegertracing.io). This phase might require us to build an efficient tracing library in Java just like https://github.com/tikv/minitrace-rust or https://github.com/tikv/minitrace-go

PR #485, opened 2022-01-19
[to #477] Change SlowLog interface to align with tracing

peng1999:
Signed-off-by: Peng Guanwen <pg999w@outlook.com>

Issue Number: to #477 
Problem Description: **See [Proposal](https://pingcap.feishu.cn/docs/doccnyjWL25a7a0M8x8MGJ2wEWc#)**

- Interfaces in `SlowLog` and `SlowLogSpan` are adjusted.
- Time measuring in `SlowLogImpl` and `SlowLogSpanImpl` are refined.
- Log outputs are adjusted to align with tracing.

### Code changes

### Check List for Tests

This PR has been tested by the at least one of the following methods:

- Unit test **TBD**

### Related changes

- NO related changes

### DIFF
--- a/src/main/java/org/tikv/common/TiConfiguration.java
+++ b/src/main/java/org/tikv/common/TiConfiguration.java
@@ -414,6 +414,12 @@ public class TiConfiguration implements Serializable {
     LEADER_AND_FOLLOWER
   }
 
+  public TiConfiguration() {
+    if (rawKVServerSlowLogFactor < 0 || rawKVServerSlowLogFactor > 1) {
+      throw new IllegalArgumentException("rawkv_server_slowlog_factor must be in range [0, 1]");
+    }
+  }
+
   public static TiConfiguration createDefault() {
     return new TiConfiguration();
   }
@@ -1004,7 +1010,7 @@ public class TiConfiguration implements Serializable {
 
   public void setRawKVServerSlowLogFactor(double rawKVServerSlowLogFactor) {
     if (rawKVServerSlowLogFactor < 0 || rawKVServerSlowLogFactor > 1) {
-      throw new IllegalArgumentException("must be in range [0, 1]");
+      throw new IllegalArgumentException("rawkv_server_slowlog_factor must be in range [0, 1]");
     }
     this.rawKVServerSlowLogFactor = rawKVServerSlowLogFactor;
   }
--- a/src/main/java/org/tikv/common/TiConfiguration.java
+++ b/src/main/java/org/tikv/common/TiConfiguration.java
@@ -369,7 +369,7 @@ public class TiConfiguration implements Serializable {
   private Integer rawKVBatchWriteSlowLogInMS =
       getIntOption(TIKV_RAWKV_BATCH_WRITE_SLOWLOG_IN_MS).orElse(null);
   private int rawKVScanSlowLogInMS = getInt(TIKV_RAWKV_SCAN_SLOWLOG_IN_MS);
-  private double rawKVServerSlowLogFactor = getDouble(TIKV_RAWKV_SERVER_SLOWLOG_FACTOR, 2.0);
+  private double rawKVServerSlowLogFactor = getDouble(TIKV_RAWKV_SERVER_SLOWLOG_FACTOR, 0.5);
 
   private boolean tlsEnable = getBoolean(TIKV_TLS_ENABLE);
   private String trustCertCollectionFile = getOption(TIKV_TRUST_CERT_COLLECTION).orElse(null);
--- a/src/main/java/org/tikv/common/TiConfiguration.java
+++ b/src/main/java/org/tikv/common/TiConfiguration.java
@@ -369,6 +369,7 @@ public class TiConfiguration implements Serializable {
   private Integer rawKVBatchWriteSlowLogInMS =
       getIntOption(TIKV_RAWKV_BATCH_WRITE_SLOWLOG_IN_MS).orElse(null);
   private int rawKVScanSlowLogInMS = getInt(TIKV_RAWKV_SCAN_SLOWLOG_IN_MS);
+  private double rawKVServerSlowLogFactor = getDouble(TIKV_RAWKV_SERVER_SLOWLOG_FACTOR, 2.0);
 
   private boolean tlsEnable = getBoolean(TIKV_TLS_ENABLE);
   private String trustCertCollectionFile = getOption(TIKV_TRUST_CERT_COLLECTION).orElse(null);
@@ -995,6 +996,14 @@ public class TiConfiguration implements Serializable {
 
   public void setRawKVScanSlowLogInMS(int rawKVScanSlowLogInMS) {
     this.rawKVScanSlowLogInMS = rawKVScanSlowLogInMS;
+  }
+
+  public double getRawKVServerSlowLogFactor() {
+    return rawKVServerSlowLogFactor;
+  }
+
+  public void setRawKVServerSlowLogFactor(double rawKVServerSlowLogFactor) {
+    this.rawKVServerSlowLogFactor = rawKVServerSlowLogFactor;
   }
 
   public boolean isCircuitBreakEnable() {
--- a/src/main/java/org/tikv/common/TiConfiguration.java
+++ b/src/main/java/org/tikv/common/TiConfiguration.java
@@ -1003,6 +1003,9 @@ public class TiConfiguration implements Serializable {
   }
 
   public void setRawKVServerSlowLogFactor(double rawKVServerSlowLogFactor) {
+    if (rawKVServerSlowLogFactor < 0 || rawKVServerSlowLogFactor > 1) {
+      throw new IllegalArgumentException("must be in range [0, 1]");
+    }
     this.rawKVServerSlowLogFactor = rawKVServerSlowLogFactor;
   }
 
--- a/src/main/java/org/tikv/common/log/SlowLogImpl.java
+++ b/src/main/java/org/tikv/common/log/SlowLogImpl.java
@@ -20,6 +20,7 @@ package org.tikv.common.log;
 import com.google.gson.JsonArray;
 import com.google.gson.JsonObject;
 import java.math.BigInteger;
+import java.text.SimpleDateFormat;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Random;
@@ -89,7 +90,11 @@ public class SlowLogImpl implements SlowLog {
     long currentNS = System.nanoTime();
     long durationMS = (currentNS - startNS) / 1_000_000;
     if (slowThresholdMS >= 0 && durationMS > slowThresholdMS) {
-      logger.warn("A slow request spent " + durationMS + "ms");
+      SimpleDateFormat dateFormat = new SimpleDateFormat("HH:mm:ss.SSS");
+      logger.warn(
+          String.format(
+              "A slow request spent %s ms. start=%s, end=%s",
+              durationMS, dateFormat.format(startMS), dateFormat.format(startMS + durationMS)));
       return true;
     }
     return false;
--- a/src/main/java/org/tikv/common/log/SlowLogImpl.java
+++ b/src/main/java/org/tikv/common/log/SlowLogImpl.java
@@ -68,6 +68,11 @@ public class SlowLogImpl implements SlowLog {
   }
 
   @Override
+  public long getThresholdMS() {
+    return slowThresholdMS;
+  }
+
+  @Override
   public void setError(Throwable err) {
     this.error = err;
   }
--- a/src/main/java/org/tikv/common/log/SlowLogImpl.java
+++ b/src/main/java/org/tikv/common/log/SlowLogImpl.java
@@ -71,7 +71,11 @@ public class SlowLogImpl implements SlowLog {
   boolean timeExceeded() {
     long currentNS = System.nanoTime();
     long durationMS = (currentNS - startNS) / 1_000_000;
-    return slowThresholdMS >= 0 && durationMS > slowThresholdMS;
+    if (slowThresholdMS >= 0 && durationMS > slowThresholdMS) {
+      logger.warn("A slow request spent " + durationMS + "ms");
+      return true;
+    }
+    return false;
   }
 
   private String getSlowLogString() {
--- a/src/main/java/org/tikv/common/log/SlowLogImpl.java
+++ b/src/main/java/org/tikv/common/log/SlowLogImpl.java
@@ -47,6 +47,8 @@ public class SlowLogImpl implements SlowLog {
 
   private final long traceId;
 
+  private long durationMS;
+
   public SlowLogImpl(long slowThresholdMS) {
     this.startMS = System.currentTimeMillis();
     this.startNS = System.nanoTime();
@@ -82,22 +84,21 @@ public class SlowLogImpl implements SlowLog {
   @Override
   public void log() {
     if (error != null || timeExceeded()) {
-      logger.warn("SlowLog:" + getSlowLogJson());
+      SimpleDateFormat dateFormat = new SimpleDateFormat("HH:mm:ss.SSS");
+      logger.warn(
+          String.format(
+              "A request spent %s ms. start=%s, end=%s, SlowLog:%s",
+              durationMS,
+              dateFormat.format(startMS),
+              dateFormat.format(startMS + durationMS),
+              getSlowLogJson().toString()));
     }
   }
 
   boolean timeExceeded() {
     long currentNS = System.nanoTime();
-    long durationMS = (currentNS - startNS) / 1_000_000;
-    if (slowThresholdMS >= 0 && durationMS > slowThresholdMS) {
-      SimpleDateFormat dateFormat = new SimpleDateFormat("HH:mm:ss.SSS");
-      logger.warn(
-          String.format(
-              "A slow request spent %s ms. start=%s, end=%s",
-              durationMS, dateFormat.format(startMS), dateFormat.format(startMS + durationMS)));
-      return true;
-    }
-    return false;
+    durationMS = (currentNS - startNS) / 1_000_000;
+    return slowThresholdMS >= 0 && durationMS > slowThresholdMS;
   }
 
   JsonObject getSlowLogJson() {
--- a/src/main/java/org/tikv/common/log/SlowLogImpl.java
+++ b/src/main/java/org/tikv/common/log/SlowLogImpl.java
@@ -21,6 +21,7 @@ import com.google.gson.JsonArray;
 import com.google.gson.JsonObject;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Random;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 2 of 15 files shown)


andylokandy:
I think we also need to fill the `trace_context` in `kvrpcpb.Context`.

src/main/java/org/tikv/common/region/AbstractRegionStoreClient.java
```diff
@@ -152,12 +154,25 @@ public boolean onStoreUnreachable(BackOffer backOffer) {
...
   }
 
-  protected Kvrpcpb.Context makeContext(TiStoreType storeType) {
-    return region.getReplicaContext(java.util.Collections.emptySet(), storeType);
+  private Kvrpcpb.Context addTraceId(Kvrpcpb.Context context, SlowLog slowLog) {
+    long traceId = slowLog.getTraceId();
+    return Kvrpcpb.Context.newBuilder(context)
+        .setTraceContext(
+            Tracepb.TraceContext.newBuilder()
+                .setDurationThresholdMs((int) slowLog.getThresholdMS())
```
andylokandy, line 162: I think we should set the threshold for tikv less than the client. Maybe half of the client threshold? @marsishandsome what do you think?

peng1999: Should we make the factor(client threshold / tikv threshold) configurable?

andylokandy: Maybe we can leave it later

andylokandy: Sounds reasonable

marsishandsome (approved):
LGTM

marsishandsome (commented):
How can users disable slowlog? Is there a switch？

src/main/java/org/tikv/common/log/SlowLogImpl.java
```diff
@@ -43,19 +43,13 @@
...
     this.startMS = System.currentTimeMillis();
     this.startNS = System.nanoTime();
     this.slowThresholdMS = slowThresholdMS;
-    this.properties = new HashMap<>(properties);
-  }
-
-  @Override
-  public void addProperty(String key, String value) {
-    this.properties.put(key, value);
+    traceId = random.nextLong();
```
zz-jason, line 52: why use random trace id? it could be conflicted with other traces in other requests, is it acceptable, why?

src/main/java/org/tikv/common/log/SlowLogImpl.java
```diff
@@ -68,45 +62,53 @@ public synchronized SlowLogSpan start(String name) {
...
+  }
+
+  boolean timeExceeded() {
     long currentNS = System.nanoTime();
-    long currentMS = startMS + (currentNS - startNS) / 1_000_000;
-    if (error != null || (slowThresholdMS >= 0 && currentMS - startMS > slowThresholdMS)) {
-      logger.warn("SlowLog:" + getSlowLogString(currentMS));
+    long durationMS = (currentNS - startNS) / 1_000_000;
+    if (slowThresholdMS >= 0 && durationMS > slowThresholdMS) {
+      logger.warn("A slow request spent " + durationMS + "ms");
```
zz-jason, line 91: it is duplicated with the log printed by the callers, is it necessary to print this log?

src/main/java/org/tikv/common/log/SlowLogImpl.java
```diff
@@ -68,45 +62,53 @@ public synchronized SlowLogSpan start(String name) {
...
     }
+    return false;
   }
 
-  private String getSlowLogString(long currentMS) {
-    SimpleDateFormat dateFormat = getSimpleDateFormat();
+  JsonObject getSlowLogJson() {
     JsonObject jsonObject = new JsonObject();
 
-    jsonObject.addProperty("start", dateFormat.format(startMS));
```
zz-jason, line 89: why remove the start and end time? how to know when a slow request is issued and when it's finished processing?

src/main/java/org/tikv/common/log/SlowLogSpanImpl.java
```diff
@@ -17,78 +17,76 @@
...
   public JsonElement toJsonElement() {
-    SimpleDateFormat simpleDateFormat = getSimpleDateFormat();
     JsonObject jsonObject = new JsonObject();
-    jsonObject.addProperty("name", name);
-    jsonObject.addProperty("start", getStartString(simpleDateFormat));
-    jsonObject.addProperty("end", getEndString(simpleDateFormat));
-    jsonObject.addProperty("duration", getDurationString());
+    jsonObject.addProperty("event", name);
+    jsonObject.addProperty("begin_unix_ns", getStartUnixNS());
+    jsonObject.addProperty("duration_ns", getDurationNS());
```
zz-jason, line 69: `ms` is more readable, why use `ns` as the time unit?

src/main/java/org/tikv/common/log/SlowLogSpanImpl.java
```diff
@@ -17,78 +17,76 @@
...
-
-  private String getStartString(SimpleDateFormat simpleDateFormat) {
-    if (startMS == 0) {
-      return "N/A";
+    if (!properties.isEmpty()) {
+      JsonObject propertiesObject = new JsonObject();
+      for (Map.Entry<String, String> entry : properties.entrySet()) {
+        propertiesObject.addProperty(entry.getKey(), entry.getValue());
+      }
+      jsonObject.add("properties", propertiesObject);
```
zz-jason, line 76: should it be put into the above `if` statement?

src/main/java/org/tikv/common/log/SlowLogSpanImpl.java
```diff
@@ -17,78 +17,76 @@
...
+  private long getStartUnixNS() {
+    return requestStartUnixNS + (startInstantNS - requestStartInstantNS);
   }
 
-  private String getDurationString() {
-    if (startMS == 0 || endMS == 0) {
-      return "N/A";
+  private long getDurationNS() {
+    if (startInstantNS == 0 || endInstantNS == 0) {
+      return 0;
```
zz-jason, line 88: how about using a negative value to indicate this scenario?

src/main/java/org/tikv/common/region/AbstractRegionStoreClient.java
```diff
@@ -152,12 +154,26 @@ public boolean onStoreUnreachable(BackOffer backOffer) {
...
 
-  protected Kvrpcpb.Context makeContext(TiStoreType storeType) {
-    return region.getReplicaContext(java.util.Collections.emptySet(), storeType);
+  private Kvrpcpb.Context addTraceId(Kvrpcpb.Context context, SlowLog slowLog) {
+    long traceId = slowLog.getTraceId();
+    return Kvrpcpb.Context.newBuilder(context)
+        .setTraceContext(
+            Tracepb.TraceContext.newBuilder()
+                .setDurationThresholdMs(
+                    (int) (slowLog.getThresholdMS() / conf.getRawKVServerSlowLogFactor()))
```
zz-jason, line 163: usually, a factor is used to be multiple with the base number.

src/main/java/org/tikv/common/log/SlowLogImpl.java
peng1999: OpenTelemetry also use random number as trace_id (See the [doc](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/api.md)). It is widely considered acceptable to rely on random numbers being unique if those numbers have enough bits. See https://softwareengineering.stackexchange.com/a/339140.

A 128-bit UUID will be better than current 64-bit random number (and is used by OpenTelemetry), but uint64 as trace_id has already used in [kvproto](https://github.com/pingcap/kvproto/commit/3fa8fa04f898c8b2f5f9b7b8576f38016cb2d5e6#diff-a546353419478c2fd5b495073d1eee706048b7524e9973f0aa846e0959c5ba42), so I just follow that.

src/main/java/org/tikv/common/log/SlowLogImpl.java
peng1999: This is aimed to print a human-friendly total time in ms, while the JSON is considered to be machine-friendly.

src/main/java/org/tikv/common/log/SlowLogImpl.java
peng1999: All the information is moved to SlowLogSpan, see the [Proposal](https://pingcap.feishu.cn/docs/doccnyjWL25a7a0M8x8MGJ2wEWc)

src/main/java/org/tikv/common/region/AbstractRegionStoreClient.java
peng1999: How about `rawkv_server_slowlog_threshold_divisor`?

src/main/java/org/tikv/common/log/SlowLogSpanImpl.java
peng1999: Which `if` statement? I think it is already in an if statement.

https://github.com/tikv/client-java/blob/f560809ff8ba78950811f0b61d22347da5644744/src/main/java/org/tikv/common/log/SlowLogSpanImpl.java#L71-L77

src/main/java/org/tikv/common/log/SlowLogSpanImpl.java
peng1999: Fixed

src/main/java/org/tikv/common/region/AbstractRegionStoreClient.java
andylokandy: I prefer multiplying the factor.

iosmanthus: Why not just use a normal tracing threshold?

peng1999: > I prefer multiplying the factor.

The parameter is now multiplying the threshold.

peng1999: > Why not just use a normal tracing threshold?

We may want to set the threshold for tikv less than the client. For example half of the client threshold.

zz-jason: it's better to limit the range of threshold factor to (0, 1) and multiple the factor in this place.

src/main/java/org/tikv/common/log/SlowLogImpl.java
```diff
@@ -68,45 +64,71 @@ public synchronized SlowLogSpan start(String name) {
...
+
+  boolean timeExceeded() {
     long currentNS = System.nanoTime();
-    long currentMS = startMS + (currentNS - startNS) / 1_000_000;
-    if (error != null || (slowThresholdMS >= 0 && currentMS - startMS > slowThresholdMS)) {
-      logger.warn("SlowLog:" + getSlowLogString(currentMS));
+    long durationMS = (currentNS - startNS) / 1_000_000;
+    if (slowThresholdMS >= 0 && durationMS > slowThresholdMS) {
+      SimpleDateFormat dateFormat = new SimpleDateFormat("HH:mm:ss.SSS");
+      logger.warn(
```
zz-jason, line 94: please remove this log, only print one log when `log()` is called.

src/main/java/org/tikv/common/log/SlowLogImpl.java
```diff
@@ -68,45 +64,71 @@ public synchronized SlowLogSpan start(String name) {
...
-    jsonObject.addProperty("end", dateFormat.format(currentMS));
-    jsonObject.addProperty("duration", (currentMS - startMS) + "ms");
     if (error != null) {
       jsonObject.addProperty("error", error.getMessage());
     }
 
-    for (Map.Entry<String, String> entry : properties.entrySet()) {
-      jsonObject.addProperty(entry.getKey(), entry.getValue());
-    }
+    jsonObject.addProperty("trace_id", toUnsignedBigInteger(traceId));
```
zz-jason, line 110: `trace_id` should come first before `error`

src/main/java/org/tikv/common/log/SlowLogImpl.java
```diff
@@ -68,45 +64,71 @@ public synchronized SlowLogSpan start(String name) {
...
     }
     jsonObject.add("spans", jsonArray);
 
-    return jsonObject.toString();
+    return jsonObject;
   }
 
-  public static SimpleDateFormat getSimpleDateFormat() {
-    return new SimpleDateFormat("HH:mm:ss.SSS");
+  /** (Copied from openjdk) Return a BigInteger equal to the unsigned value of the argument. */
```
zz-jason, line 121: open jdk is under the [GPL v2](https://github.com/openjdk/jdk/blob/master/LICENSE) license, if we copy the code, client-java should be open-sourced as GPL v2 as well if I understand correctly. Please double-check it.

src/main/java/org/tikv/common/log/SlowLogSpanImpl.java
```diff
@@ -17,78 +17,76 @@
...
   @Override
   public JsonElement toJsonElement() {
-    SimpleDateFormat simpleDateFormat = getSimpleDateFormat();
     JsonObject jsonObject = new JsonObject();
-    jsonObject.addProperty("name", name);
-    jsonObject.addProperty("start", getStartString(simpleDateFormat));
-    jsonObject.addProperty("end", getEndString(simpleDateFormat));
-    jsonObject.addProperty("duration", getDurationString());
+    jsonObject.addProperty("event", name);
+    jsonObject.addProperty("begin_unix_ms", getStartUnixNS() / 1_000_000);
```
zz-jason, line 68: Please print human-readable timestamps, you can use `SimpleDateFormat`. It's better to change the field name to `begin`.

src/main/java/org/tikv/common/log/SlowLogSpanImpl.java
```diff
@@ -17,78 +17,76 @@
...
+  private long getStartUnixNS() {
+    return requestStartUnixNS + (startInstantNS - requestStartInstantNS);
   }
 
-  private String getDurationString() {
-    if (startMS == 0 || endMS == 0) {
-      return "N/A";
+  private long getDurationNS() {
+    if (startInstantNS == 0 || endInstantNS == 0) {
+      return -1;
```
zz-jason, line 88: still confused about why not return `N/A`?

src/main/java/org/tikv/common/log/SlowLogImpl.java
zz-jason: we should limit the log amount as less as possible. please only print one slowlog for one request.

src/main/java/org/tikv/common/log/SlowLogSpanImpl.java
peng1999: `"N/A"` will make the field have union type `string | number`, which is not good for parse.

src/main/java/org/tikv/common/log/SlowLogImpl.java
peng1999: Fixed.

src/main/java/org/tikv/common/log/SlowLogImpl.java
peng1999: Fixed. The tow logs are merged.

src/main/java/org/tikv/common/log/SlowLogSpanImpl.java
peng1999: Added in 5be5f0b

src/main/java/org/tikv/common/log/SlowLogImpl.java
peng1999: The code is rewritten in d8ff43a

src/main/java/org/tikv/common/TiConfiguration.java
```diff
@@ -997,6 +998,14 @@ public void setRawKVScanSlowLogInMS(int rawKVScanSlowLogInMS) {
     this.rawKVScanSlowLogInMS = rawKVScanSlowLogInMS;
   }
 
+  public double getRawKVServerSlowLogFactor() {
+    return rawKVServerSlowLogFactor;
+  }
+
+  public void setRawKVServerSlowLogFactor(double rawKVServerSlowLogFactor) {
+    this.rawKVServerSlowLogFactor = rawKVServerSlowLogFactor;
```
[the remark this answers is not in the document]
peng1999: Fixed in c5ba12b and 38e8a7a

src/main/java/org/tikv/common/log/SlowLogSpanImpl.java
zz-jason: what is `begin_unix_ms` used for, seems it can be infered from `begin`, can we remove it?

src/main/java/org/tikv/common/log/SlowLogSpanImpl.java
```diff
@@ -17,78 +17,79 @@
...
+    SimpleDateFormat dateFormat = new SimpleDateFormat("HH:mm:ss.SSS");
     JsonObject jsonObject = new JsonObject();
-    jsonObject.addProperty("name", name);
-    jsonObject.addProperty("start", getStartString(simpleDateFormat));
-    jsonObject.addProperty("end", getEndString(simpleDateFormat));
-    jsonObject.addProperty("duration", getDurationString());
+    jsonObject.addProperty("event", name);
+    jsonObject.addProperty("begin", dateFormat.format(getStartUnixNS() / 1_000_000));
+    jsonObject.addProperty("begin_unix_ms", getStartUnixNS() / 1_000_000);
+    jsonObject.addProperty("duration_ms", getDurationNS() / 1_000_000);
```
zz-jason, line 72: would it result in a `long` value since both values are integers?

src/main/java/org/tikv/common/log/SlowLogImpl.java
```diff
@@ -68,45 +66,66 @@ public synchronized SlowLogSpan start(String name) {
...
     this.error = err;
   }
 
   @Override
   public void log() {
-    long currentNS = System.nanoTime();
-    long currentMS = startMS + (currentNS - startNS) / 1_000_000;
-    if (error != null || (slowThresholdMS >= 0 && currentMS - startMS > slowThresholdMS)) {
-      logger.warn("SlowLog:" + getSlowLogString(currentMS));
+    if (error != null || timeExceeded()) {
```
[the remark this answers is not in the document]
peng1999: If an exception is thrown and not caught, the error message will be printed by java runtime and included in the `error` field of the slow log JSON.

src/main/java/org/tikv/common/log/SlowLogSpanImpl.java
peng1999: begin_unix_ms is removed.

zz-jason (approved):
LGTM

ti-srebot (approved):

zz-jason (approved):

iosmanthus (approved):
LGTM

PR #496, opened 2022-01-25
[close #497] let JVM ClassLoader load gRPC error related classes during warmup

### LINKED ISSUE #497: request timeout when receiving NOT_LEADER error the first time
Request timeout when receiving NOT_LEADER error the first time.

The reason is that [error.toString](https://github.com/tikv/client-java/blob/v3.1.9/src/main/java/org/tikv/common/operation/RegionErrorHandler.java#L95) will trigger JVM ClassLoadr load grpc error related classes, which costs more than 100ms.

marsishandsome:

close https://github.com/tikv/client-java/issues/497

Problem Description:  Request timeout when receiving NOT_LEADER error the first time.

The reason is that [error.toString](https://github.com/tikv/client-java/blob/v3.1.9/src/main/java/org/tikv/common/operation/RegionErrorHandler.java#L95) will trigger JVM ClassLoader load grpc error related classes, which costs more than 100ms.

Solution: let JVM JVM ClassLoader load grpc error related classes during warmup.

### DIFF
--- a/src/main/java/org/tikv/common/TiSession.java
+++ b/src/main/java/org/tikv/common/TiSession.java
@@ -36,6 +36,7 @@ import org.tikv.common.key.Key;
 import org.tikv.common.meta.TiTimestamp;
 import org.tikv.common.region.*;
 import org.tikv.common.util.*;
+import org.tikv.kvproto.Errorpb;
 import org.tikv.kvproto.ImportSstpb;
 import org.tikv.kvproto.Metapb;
 import org.tikv.kvproto.Pdpb;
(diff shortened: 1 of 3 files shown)


zz-jason (commented):
could you add some test?

marsishandsome:
> could you add some test?

sure.

src/main/java/org/tikv/common/TiSession.java
```diff
@@ -166,6 +167,13 @@ private synchronized void warmUp() {
     long warmUpStartTime = System.nanoTime();
     BackOffer backOffer = ConcreteBackOffer.newRawKVBackOff();
     try {
+      // let JVM ClassLoader load gRPC error related classes
+      // this operation may cost 100ms
+      Errorpb.Error.newBuilder()
+          .setNotLeader(Errorpb.NotLeader.newBuilder().build())
```
zz-jason, line 173: do we need to warm up other errors?

marsishandsome: All the classes (include other errors) in `org.tikv.kvproto.Errorpb` will be loaded during warmup.

zz-jason (approved):
LGTM

PR #516, opened 2022-02-14
[close #515] Slowlog: Always calculate duration

### LINKED ISSUE #515: Slowlog does not record total duration time when error happend
### 1. Describe the bug

Slowlog does not record total duration time when error happend

print a slowlog that contains error.

zero duration is printed in log.

print the real duration time in log

### 5. What is your Java Client and TiKV version? (Required)

- Client Java: 3.2.0-SNAPSHOT
- TiKV: nightly

peng1999:
Signed-off-by: Peng Guanwen <pg999w@outlook.com>

Issue Number: close #515

Problem Description: Always calculate duration in slowlog

### Check List for Tests

This PR has been tested by the at least one of the following methods:
- Unit test

### Side effects

- NO side effects

### Related changes

- NO related changes

### DIFF
--- a/src/main/java/org/tikv/common/log/SlowLogImpl.java
+++ b/src/main/java/org/tikv/common/log/SlowLogImpl.java
@@ -83,7 +83,8 @@ public class SlowLogImpl implements SlowLog {
 
   @Override
   public void log() {
-    if (error != null || timeExceeded()) {
+    boolean timeExceeded = recordTime();
+    if (error != null || timeExceeded) {
       SimpleDateFormat dateFormat = new SimpleDateFormat("HH:mm:ss.SSS");
       logger.warn(
           String.format(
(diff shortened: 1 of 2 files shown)


zz-jason (approved):
LGTM

src/main/java/org/tikv/common/log/SlowLogImpl.java
```diff
@@ -95,7 +96,7 @@ public void log() {
     }
   }
 
-  boolean timeExceeded() {
+  boolean recordTime() {
```
iosmanthus, line 99: This name confuses me. Could we change it to `ifRecordTime`?

iosmanthus (approved):
LGTM

src/main/java/org/tikv/common/log/SlowLogImpl.java
andylokandy, line 99: maybe it's better to split it into two methods...

andylokandy, line 99: maybe it's better to split it into two methods...

andylokandy (changes requested):

andylokandy (approved):

PR #518, opened 2022-02-14
[to #439] introduce MockThreeStoresTest to onStoreUnreachable

### LINKED ISSUE #439: Using MockTiKV to do unit test
Currently, we use a real TiKV instance to do unit test. 

It's very hard to test errors in such mode, e.g. https://github.com/tikv/client-java/pull/434.

We should use MockTiKV to test errors.

iosmanthus:

Issue Number: close #439

This pull request tries to mock a TiKV grpc server and cover the error handling logic of client-java, mostly, the logic of `seekLeaderStore` and `seekProxyStore` in `handleRequestError`.


src/main/java/org/tikv/common/TiSession.java
```diff
@@ -166,6 +178,10 @@ private static VersionInfo getVersionInfo() {
   }
 
   private synchronized void warmUp() {
+    if (conf.isTest()) {
+      logger.info("skip warm up in test mode");
```
zz-jason, line 182: why skip warmup in test mode?

src/test/java/org/tikv/common/KVMockServer.java
```diff
@@ -97,7 +124,7 @@ private void verifyContext(Context context) throws Exception {
     if (context.getRegionId() != region.getId()
         || !context.getRegionEpoch().equals(region.getRegionEpoch())
         || !context.getPeer().equals(region.getLeader())) {
-      throw new Exception();
+      throw new Exception("context doesn't match");
```
zz-jason, line 127: it would be better to specifically point out the mismatched field in context.

src/main/java/org/tikv/common/TiSession.java
iosmanthus: 1. `PDMockServerTest` is the base class of `MockServerTest`.
2. `TiSession` is initialized in `PDMockServerTest`, and it calls `warmup`.
3. `warmup` calls `getAllStores` and  `rawGet`.

This results in a dependent hell. We should ignore this anti-pattern warmup operation in `TiSession`, otherwise, we could not launch `MockServerTest` propperly.

src/main/java/org/tikv/common/region/AbstractRegionStoreClient.java
```diff
@@ -230,6 +231,7 @@ private Boolean seekLeaderStore(BackOffer backOffer) {
             // switch to leader store
             store = currentLeaderStore;
             updateClientStub();
+            return true;
```
marsishandsome, line 234: you fix a bug here?

src/main/java/org/tikv/common/region/TiRegion.java
```diff
@@ -40,6 +40,7 @@
 import org.tikv.kvproto.Metapb.Region;
 
 public class TiRegion implements Serializable {
+
```
marsishandsome, line 43: you can do code format in another PR

src/test/java/org/tikv/common/KVMockServer.java
```diff
@@ -64,10 +78,23 @@
   public static final int STORE_NOT_MATCH = 9;
   public static final int RAFT_ENTRY_TOO_LARGE = 10;
 
+  public enum State {
+    Normal,
+    Fail
```
marsishandsome, line 83: `Fail` means the server is down?

src/test/java/org/tikv/common/MockThreeStoresTest.java
```diff
@@ -0,0 +1,129 @@
...
+import java.util.List;
+import java.util.stream.Collectors;
+import org.junit.After;
+import org.junit.Before;
+import org.tikv.common.region.TiRegion;
+import org.tikv.common.region.TiStore;
+import org.tikv.kvproto.Metapb;
+import org.tikv.kvproto.Pdpb;
+
+public class MockThreeStoresTest extends PDMockServerTest {
```
marsishandsome, line 34: rename `MockThreeStoresTest` to `MockTiKVClusterTest`?
The store number can be config?

src/main/java/org/tikv/common/region/AbstractRegionStoreClient.java
iosmanthus: Yes. If it returns false here, the region info will update for nothing.

src/test/java/org/tikv/common/KVMockServer.java
iosmanthus: Yes.

src/test/java/org/tikv/common/MockThreeStoresTest.java
iosmanthus: I just want to keep it simple enough for now, refactoring will be introduce in the future.

src/main/java/org/tikv/common/region/AbstractRegionStoreClient.java
marsishandsome: Is there a testcase to verify this fix?

src/test/java/org/tikv/common/MockThreeStoresTest.java
marsishandsome: Create an issue to describe this refactoring?

src/main/java/org/tikv/common/region/AbstractRegionStoreClient.java
iosmanthus: Yes. `SeekLeaderStore.testSeekLeader` cover this case. If return false, the tests will fail.

src/main/java/org/tikv/common/TiSession.java
iosmanthus: Ignored warmup by `setEnableWarmup(false)`

src/test/java/org/tikv/common/MockThreeStoresTest.java
marsishandsome: sure

marsishandsome (approved):
LGTM

src/main/java/org/tikv/common/region/AbstractRegionStoreClient.java
zz-jason: should it be cherry-picked to release-3.1?

zz-jason (approved):
LGTM

ti-srebot (approved):

Issue #556, opened 2022-03-16
Attach cluster info to slow log and metrics


iosmanthus:
## Feature Request

**Is your feature request related to a problem? Please describe:**

While handling multiple clusters with a single client, we need some information to distinguish them in metrics and slow log outputs. Otherwise, it's hard to tell which cluster is dubious. 

**Describe the feature you'd like:**
- [ ] Attach `cluster_id` and `pd_addresses` properties to slow log.
- [ ] Attach `cluster_id` label to metrics.

PR #558, opened 2022-03-18
[to #556] metrics: attach cluster label to metrics

iosmanthus:

Issue Number: to #556

Problem Description:

- [x] code change for client-java
- [x] code change for grafana panel

Add `cluster` label for metrics in client-java with the value of cluster id.


src/main/java/org/tikv/common/ClusterIdRecorder.java
```diff
@@ -0,0 +1,32 @@
...
+ */
+
+package org.tikv.common;
+
+import java.util.concurrent.atomic.AtomicLong;
+
+public interface ClusterIdRecorder {
+  AtomicLong clusterId = new AtomicLong(0);
+
+  default void withClusterId(long id) {
```
zz-jason, line 25: would it better to be renamed to `setClustrerId()`?

zz-jason (commented):
since each session is connected to a tikv cluster, can you init the cluster id when `TiSession` is created?

marsishandsome (commented):
LGTM

marsishandsome (approved):

src/main/java/org/tikv/common/PDClient.java
```diff
@@ -347,6 +354,7 @@ private boolean isScatterRegionFinish(GetOperatorResponse resp) {
   @Override
   public List<Pdpb.Region> scanRegions(
       BackOffer backOffer, ByteString startKey, ByteString endKey, int limit) {
+    backOffer.withClusterId(getClusterId());
```
zz-jason, line 357: is it possible to obtain cluster id when the `backoffer` is created?

src/main/java/org/tikv/common/util/ConcreteBackOffer.java
```diff
@@ -76,37 +79,54 @@ private ConcreteBackOffer(ConcreteBackOffer source) {
...
-  public static ConcreteBackOffer newDeadlineBackOff(int timeoutInMs, SlowLog slowLog) {
+  public static ConcreteBackOffer newDeadlineBackOff(
+      int timeoutInMs, SlowLog slowLog, long clusterId) {
     long deadline = System.currentTimeMillis() + timeoutInMs;
-    return new ConcreteBackOffer(0, deadline, slowLog);
+    return new ConcreteBackOffer(0, deadline, slowLog, clusterId);
+  }
+
+  public static ConcreteBackOffer newDeadlineBackOff(int timeoutInMs, SlowLog slowLog) {
+    return newDeadlineBackOff(timeoutInMs, slowLog, 0);
```
zz-jason, line 89: seems this function is never called? otherwise, the cluster id is incorrect.

src/main/java/org/tikv/common/util/ConcreteBackOffer.java
```diff
@@ -76,37 +79,54 @@ private ConcreteBackOffer(ConcreteBackOffer source) {
...
   }
 
   public static ConcreteBackOffer newCustomBackOff(int maxSleep) {
-    return new ConcreteBackOffer(maxSleep, 0, SlowLogEmptyImpl.INSTANCE);
+    return newCustomBackOff(maxSleep, 0);
   }
 
   public static ConcreteBackOffer newScannerNextMaxBackOff() {
-    return new ConcreteBackOffer(SCANNER_NEXT_MAX_BACKOFF, 0, SlowLogEmptyImpl.INSTANCE);
+    return new ConcreteBackOffer(SCANNER_NEXT_MAX_BACKOFF, 0, SlowLogEmptyImpl.INSTANCE, 0);
```
zz-jason, line 101: ditto

src/main/java/org/tikv/common/util/ConcreteBackOffer.java
iosmanthus: It is used in tests: https://github.com/iosmanthus/client-java/blob/e85ae32c5a232be34c3819d3a0fd1050aca3311d/src/test/java/org/tikv/raw/RawKVClientTest.java#L222. Some of the cluster id is irrelevant in test, so it is set to 0.

zz-jason: can we delete it if it's only used in tests?

zz-jason (approved):
LGTM

ti-srebot (approved):

PR #575, opened 2022-03-24
[to #555] enable api v2

### LINKED ISSUE #555: Support TiKV API V2
## Feature Request

**Describe the feature you'd like:**

Support TiKV API V2, to obtain more TiKV features such as ["cross cluster replication"](https://github.com/tikv/rfcs/pull/86), mixed using TiDB + TxnKV + RawKV on the same TiKV cluster, etc.
See [API V2 proposal](https://github.com/tikv/rfcs/blob/master/text/0069-api-v2.md) and [its implementation details](https://github.com/pingyu/tikv-rfcs/blob/rfc_api_v2_pd/text/0069-api-v2.md) for detail.
Refer to https://github.com/tikv/tikv/issues/10938.

**Teachability, Documentation, Adoption, Migration Strategy:**

Some developing features (including "Cross Cluster Replication", mixed using TiDB + TxnKV + RawKV) of TiKV depends on API V2.
We also recommend that users switch to API V2 ASAP, as most features in the future will also be based on API V2.
Data migration is required, as underlying data encoding between API V2 and the early version is quite different. Tools for migration will be provided in the near future.

**Modification Required:**

1. Set `api_version` = 2 in [`kvrpcpb.Context`](https://github.com/pingcap/kvproto/blob/10e6883072219d03234f767d5fe994edb44ba366/proto/kvrpcpb.proto#L711).
2. Prefix requesting keys by `r` for RawKV, and `x` for TxnKV.
3. Request region info from PD with keys having `r/x` prefix and Mem-comparable Padding.
4. Remove prefix of keys from the response if any (e.g. `raw_scan`, `scan`).

iosmanthus:

- [x] attach API version to context to `AbstractRegionStoreClient`, `RegionStoreClient`, `LockResolverClientV4`.
- [x] simple API conf check.

Issue Number: close #555 , and tracked by https://github.com/tikv/client-java/issues/582

### Check List for Tests

- [x] Unit test
	- [x] RawKV compatibility
	- [ ] TxnKV compatibility
- [x] Integration test

### Related changes

- Need to cherry-pick the release branch: **v3.3**


src/main/java/org/tikv/common/region/AbstractRegionStoreClient.java
```diff
@@ -155,6 +156,42 @@ public boolean onStoreUnreachable(BackOffer backOffer) {
     return false;
   }
 
+  public ByteString buildRequestKey(ByteString key) {
+    switch (conf.getApiVersion()) {
```
zz-jason, line 160: can we get the cluster version and tikv configuration from PD? so that we don't need to add one more config in the java client and the java client can automatically adjust the encode/decode methods according to the cluster it is connected to.

iosmanthus: By design, the client should not read the TiKV configuration from PD or themself. Because:

1. TiKV configurations could be different from each other due to the restart process.
2. TiDB use client-go, which should have the same behaviour as the client-java, however, TiDB would never use API v2 but use API v1 instead, due to they have a key prefix with `m` or `t`. Thus, client-go should not read the API version from the cluster components, instead, just set the version to `1`.

src/test/java/org/tikv/common/ApiVersionTest.java
```diff
@@ -0,0 +1,153 @@
...
+import java.util.Optional;
+import org.junit.Assert;
+import org.junit.Assume;
+import org.junit.Test;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.tikv.common.TiConfiguration.ApiVersion;
+import org.tikv.raw.RawKVClient;
+
+public class ApiVersionTest {
```
iosmanthus, line 30: @pingyu PTAL on this compatible test for API v2.

pingyu: LGTM~

src/main/java/org/tikv/common/ConfigUtils.java
```diff
@@ -200,4 +203,6 @@ public class ConfigUtils {
   public static final int DEF_TiKV_CIRCUIT_BREAK_ATTEMPT_REQUEST_COUNT = 10;
 
   public static final int DEF_TIKV_SCAN_REGIONS_LIMIT = 1000;
+
+  public static final int DEF_TIKV_API_VERSION = 1;
```
zz-jason, line 207: we have three options: v1, v1ttl, and v2, better to use a string?

src/main/java/org/tikv/common/TiConfiguration.java
```diff
@@ -1075,4 +1223,103 @@ public int getScanRegionsLimit() {
...
+
+    public boolean isV1() {
+      return this == V1;
+    }
+
+    public boolean isV2() {
+      return this == V2;
+    }
+
+    public Kvrpcpb.APIVersion toPb() {
```
zz-jason, line 1314: should v1ttl also be considered?

zz-jason, line 1314: should v1ttl also be considered?

xuanyu66:
@iosmanthus plz fix the CI

src/main/java/org/tikv/common/PDClient.java
```diff
@@ -811,43 +819,60 @@ public String toString() {
...
+  public Metapb.Region decodeRegion(Metapb.Region region) {
     final boolean isRawRegion = conf.isRawKVMode();
     Metapb.Region.Builder builder =
         Metapb.Region.newBuilder()
             .setId(region.getId())
             .setRegionEpoch(region.getRegionEpoch())
             .addAllPeers(region.getPeersList());
 
-    if (region.getStartKey().isEmpty() || isRawRegion) {
+    if (region.getStartKey().isEmpty() || (conf.getApiVersion().isV1() && isRawRegion)) {
```
sunxiaoguang, line 831: We should have a fast path that simply return original region when it's unnecessary.

src/main/java/org/tikv/common/TiConfiguration.java
```diff
@@ -1241,4 +1250,103 @@ public int getScanRegionsLimit() {
...
+      switch (this) {
+        case V1:
+          return Kvrpcpb.APIVersion.V1;
+        case V2:
+          return Kvrpcpb.APIVersion.V2;
```
sunxiaoguang, line 1346: I would suggest create two implementations instead of using many ifelses or switches.

src/main/java/org/tikv/common/region/RegionStoreClient.java
```diff
@@ -587,15 +591,15 @@ private boolean isTxnHeartBeatSuccess(TxnHeartBeatResponse resp)
    * @param startTs start version
    * @param commitTs commit version
    */
-  public void commit(BackOffer backOffer, Iterable<ByteString> keys, long startTs, long commitTs)
+  public void commit(BackOffer backOffer, List<ByteString> keys, long startTs, long commitTs)
```
sunxiaoguang, line 594: Why not convert iterable to stream directly?

sunxiaoguang: Or change the type of keys to stream directly to avoid creating redundant concrete List.

src/main/java/org/tikv/common/region/RegionStoreClient.java
```diff
@@ -1265,7 +1270,15 @@ private List<KvPair> rawScanHelper(RawScanResponse resp) {
     if (resp.hasRegionError()) {
       throw new RegionException(resp.getRegionError());
     }
-    return resp.getKvsList();
+    return resp.getKvsList()
```
sunxiaoguang, line 1273: There should be a fast path when KvsList doesn't need to be converted as well.

src/test/java/org/tikv/common/StoreConfig.java
```diff
@@ -0,0 +1,72 @@
...
+        return new Gson().fromJson(content, JsonObject.class);
+      }
+    } catch (Exception e) {
+      logger.error("Failed to get store api version: ", e);
+      throw new IllegalStateException(e);
+    }
+  }
+
+  public static ApiVersion acquireApiVersion(PDClient client) {
+    JsonElement version = getConfig(client).get("storage").getAsJsonObject().get("api-version");
```
sunxiaoguang, line 64: Are we sure the parsed config do have storage field? If not, we might see null pointer exception.

src/main/java/org/tikv/common/PDClient.java
```diff
@@ -811,43 +819,60 @@ public String toString() {
     }
   }
 
-  private Metapb.Region decodeRegion(Metapb.Region region) {
+  @Override
+  public Metapb.Region decodeRegion(Metapb.Region region) {
```
xuanyu66, line 823: From line 367, it seems like we only don't need to do anything when API is v1, but why do we need decodeRegion now when API is v1. Is there a bug in old code?

src/main/java/org/tikv/common/PDClient.java
```diff
@@ -811,43 +819,60 @@ public String toString() {
...
-      } else {
-        try {
-          byte[] decodedStartKey = BytesCodec.readBytes(new CodecDataInput(region.getStartKey()));
-          builder.setStartKey(ByteString.copyFrom(decodedStartKey));
-        } catch (Exception e) {
-          builder.setStartKey(region.getStartKey());
+      try {
+        ByteString decodedStartKey =
+            ByteString.copyFrom(BytesCodec.readBytes(new CodecDataInput(region.getStartKey())));
+        if (conf.getApiVersion().isV2()) {
```
xuanyu66, line 837: This method has so many if, can you explain in code comment what behavior do you want.
Otherwise, I don't know if where is something wrong

src/main/java/org/tikv/common/TiConfiguration.java
```diff
@@ -1241,4 +1250,103 @@ public int getScanRegionsLimit() {
...
+        }
+        // TODO: empty key is not always scan back!!!!
+        if (!key.isEmpty() && !key.startsWith(getKeyPrefix())) {
+          throw new IllegalArgumentException("key corrupted, wrong prefix");
+        }
+        return key.substring(key.isEmpty() ? 0 : 1);
+      default:
+        throw new IllegalArgumentException("unknown api version or kv mode");
+    }
+  }
```
zhangyangyu, line 1316: These methods are handling message payload. Putting them in configuration seems weird.

src/main/java/org/tikv/common/PDClient.java
xuanyu66: So we need a fast path to return when v1

iosmanthus: Has changed to using `RequestKeyCodec` to do this.

src/main/java/org/tikv/common/TiConfiguration.java
iosmanthus: V1TTL is a state that only exists in the TiKV.

src/main/java/org/tikv/common/apiversion/CodecUtils.java
```diff
@@ -0,0 +1,18 @@
...
+
+import com.google.protobuf.ByteString;
+import org.tikv.common.codec.Codec.BytesCodec;
+import org.tikv.common.codec.CodecDataInput;
+import org.tikv.common.codec.CodecDataOutput;
+
+public class CodecUtils {
+  public static ByteString encode(ByteString key) {
+    CodecDataOutput cdo = new CodecDataOutput();
+    BytesCodec.writeBytes(cdo, key.toByteArray());
```
sunxiaoguang, line 11: we can have a fast path to take internal array out without copying when implementation of ByteString is actually ByteString.LiteralByteString or ByteString.BoundedByteString. 

To use package level protected method or classes, you can either create a utility class in our source tree within com.google.protobuf package or use reflection to make them accessible.

zhangyangyu (commented):
Much better! One thing I am curious is the usage of `Pair`. There are many places using it, like `encodeRange`. Why not directly use the `Range` type?

iosmanthus:
> 

The `Range` type will use the default comparator for the elements, it tends to be an actual range that use to compute if an element is between the `min` and `max`.  The `Pair` type tells us that the return value is a tuple, nothing more.

src/main/java/org/tikv/common/TiConfiguration.java
iosmanthus: Removed. The same logic is implemented in TiSession's `RequestKeyCodec` dispatching.

xuanyu66 (approved):

src/main/java/org/tikv/common/PDClient.java
```diff
@@ -353,18 +354,21 @@ public List<Pdpb.Region> scanRegions(
     // introduce a warm-up timeout for ScanRegions requests
     PDGrpc.PDBlockingStub stub =
         getBlockingStub().withDeadlineAfter(conf.getWarmUpTimeout(), TimeUnit.MILLISECONDS);
+    org.apache.commons.lang3.tuple.Pair<ByteString, ByteString> range =
```
zhangyangyu, line 357: Is it possible to normalize the `Pair` class we use? I don't get the point to use `org.apache.commons.lang3.tuple.Pair` instead of `org.tikv.common.util.Pair` or vice versa.

src/main/java/org/tikv/common/TiSession.java
```diff
@@ -154,16 +174,20 @@ public TiSession(TiConfiguration conf) {
...
 
-    this.client = PDClient.createRaw(conf, channelFactory);
+    this.client = PDClient.createRaw(conf, keyCodec, channelFactory);
     this.enableGrpcForward = conf.getEnableGrpcForward();
     if (this.enableGrpcForward) {
       logger.info("enable grpc forward for high available");
     }
     if (conf.isWarmUpEnable() && conf.isRawKVMode()) {
-      warmUp();
+      warmup();
```
zhangyangyu, line 183: Could we keep the camel style? In TiConfiguration, the variables of warm-up are all camel style.

src/main/java/org/tikv/common/PDClient.java
iosmanthus: OK.

zhangyangyu (approved):

PR #597, opened 2022-04-22
[close #596] fix tls reloading encouter slow thread npe

### LINKED ISSUE #596: TLS reloading encounter NPE while testing with multithread
### 1. Describe the bug

TLS reloading will produce NPE while `getChannel` with multiple threads.

```java
    // Test getChannel concurrently
    factory = new ChannelFactory(v, v, v, v, caPath, clientCertPath, clientKeyPath);

    List<Thread> tasks = new ArrayList<>(8);
    for (int i = 0; i < 8; i++) {
      ChannelFactory finalFactory = factory;
      Thread t = new Thread(() -> finalFactory.getChannel("127.0.0.1:2379", mapping));
      t.start();
      tasks.add(t);
    }

    for (Thread t : tasks) {
      t.join();
    }
```

Panic

Test pass

- Client Java: https://github.com/tikv/client-java/commit/7a123a07e2c8d2498610e10dda4de9c537bcc3cc

iosmanthus:

Issue Number: close #596

Problem Description:

`getChannel` will encounter NPE for the slow thread might meet a null if the cert is unmodified.

Add an initial state for `SslContextBuilder`

### Check List for Tests

- Unit test

### Related changes


src/main/java/org/tikv/common/util/ChannelFactory.java
```diff
@@ -213,14 +214,15 @@ public ChannelFactory(
     this.keepaliveTimeout = keepaliveTimeout;
     this.idleTimeout = idleTimeout;
     this.certContext = new JksContext(jksKeyPath, jksKeyPassword, jksTrustPath, jksTrustPassword);
+    reloadSslContext();
   }
 
   @VisibleForTesting
   public boolean reloadSslContext() {
     if (certContext != null) {
       SslContextBuilder newBuilder = certContext.reload();
```
sunxiaoguang, line 223: There is still race between `isModified` check and rest of initialization code.

sunxiaoguang, line 223: There is still a race between `isModified` check and the rest of initialization logic.

sunxiaoguang: Reload in constructor can help mitigating NPE. But the race can make later certificate reload not working as expected.

iosmanthus:
@sunxiaoguang PTAL

zhangyangyu (commented):
Currently, all concurrency control applies to "getting the latest SSLContext builder"  and then "use it to initialize the connection pool correctly". But what if there are two consecutive cert change happening very quickly? Then we might use two different SSLContext builder to initialize the pool since when the second `reload` calls `connpool.clear`, the first call to `getChannel` migth stil not start initialize the pool.

iosmanthus:
> Currently, all concurrency control applies to "getting the latest SSLContext builder" and then "use it to initialize the connection pool correctly". But what if there are two consecutive cert change happening very quickly? Then we might use two different SSLContext builder to initialize the pool since when the second `reload` calls `connpool.clear`, the first call to `getChannel` might still not start to initialize the pool.

the first thread will fail while the connection user uses it, and then a retry is expected.

src/main/java/org/tikv/common/util/ChannelFactory.java
```diff
@@ -215,63 +216,69 @@ public ChannelFactory(
...
+
+    if (sslContextBuilder == null) {
+      return builder.usePlaintext().build();
+    } else {
+      SslContext sslContext;
+      try {
+        sslContext = sslContextBuilder.build();
+      } catch (SSLException e) {
+        logger.error("create ssl context failed!", e);
+        return null;
```
zhangyangyu, line 251: `return null` or `raise exception`? What should the user expect and how they could restore?

src/main/java/org/tikv/common/util/ChannelFactory.java
```diff
@@ -215,63 +216,69 @@ public ChannelFactory(
...
-      logger.info("invalidate connection pool");
+  @VisibleForTesting
+  public synchronized ManagedChannel reload(String address, HostMapping mapping) {
+    assert certContext != null;
+    // TODO: use WatchService to detect file change.
+    if (certContext.isModified()) {
+      logger.info("certificate file changed, reloading");
+      sslContextBuilder.set(certContext.createSslContextBuilder());
       connPool.clear();
+      return connPool.computeIfAbsent(
```
xuanyu66, line 265: It's more reasonable to make the method `reload` void, there is already logic in getChannel() that creates a channel. It can make code more readable

zhangyangyu, line 265: Maybe we could add a comment here explaining why we double the creation logic in case anyone in future think it's redundant and delete it :-(

src/main/java/org/tikv/common/util/ChannelFactory.java
```diff
@@ -215,63 +216,69 @@ public ChannelFactory(
...
+      return builder.sslContext(sslContext).build();
     }
-    return false;
   }
 
-  public ManagedChannel getChannel(String addressStr, HostMapping hostMapping) {
-    if (reloadSslContext()) {
-      logger.info("invalidate connection pool");
+  @VisibleForTesting
+  public synchronized ManagedChannel reload(String address, HostMapping mapping) {
```
xuanyu66, line 258: I am worried about the performance of getChannel(). Since renewing cert is a very low frequent operation, it's a big cost to sync and invoke a disk IO every time

xuanyu66: And from my point of view, client-java needs to have a high throughput when working as a cache client.
It's a situation quite like service discovery in RPC client, in which we wouldn't refresh every time getChannel

iosmanthus: > 

There is an optimization to avoid sync disk IO, using a `WatchService` to notify the `ChannelFactory` if the cert is changed. A TODO  is added.

src/main/java/org/tikv/common/util/ChannelFactory.java
iosmanthus: Done.

src/main/java/org/tikv/common/util/ChannelFactory.java
iosmanthus: change to `raise exception`.

src/main/java/org/tikv/common/util/ChannelFactory.java
xuanyu66: Would you enhance it recently?

src/main/java/org/tikv/common/util/ChannelFactory.java
```diff
@@ -26,80 +26,160 @@
...
-        logger.info("reload ssl context");
-        return createSslContextBuilder();
+    private void watchLoop() {
+      try {
+        WatchService watchService = FileSystems.getDefault().newWatchService();
+        Path p = Paths.get(path);
+        p.register(watchService, StandardWatchEventKinds.ENTRY_MODIFY);
+
+        WatchKey key;
+        while ((key = watchService.take()) != null) {
```
xuanyu66, line 101: Under which condition would equal null?

src/main/java/org/tikv/common/util/ChannelFactory.java
```diff
@@ -26,80 +26,160 @@
...
+          boolean changed = false;
+          OUTER:
+          for (WatchEvent<?> event : key.pollEvents()) {
+            for (String target : targets) {
+              logger.info("detected file change: {}", event.context());
+              if (event.context().toString().equals(target)) {
+                changed = true;
+                break OUTER;
```
xuanyu66, line 109: Why break here?

src/main/java/org/tikv/common/util/ChannelFactory.java
```diff
@@ -26,80 +26,160 @@
...
+  private int connRecycleTime;
   private final int maxFrameSize;
   private final int keepaliveTime;
   private final int keepaliveTimeout;
   private final int idleTimeout;
-  private final ConcurrentHashMap<String, ManagedChannel> connPool = new ConcurrentHashMap<>();
   private final CertContext certContext;
+
+  @VisibleForTesting
+  public final ConcurrentHashMap<Pair<SslContextBuilder, String>, ManagedChannel> connPool =
```
xuanyu66, line 70: The key of Map doesn't implement hashCode() and equals(). Will it bring some potential bugs?

xuanyu66 (changes requested):

xuanyu66:
@shiyuhang0 PTAL

src/main/java/org/tikv/common/util/ChannelFactory.java
iosmanthus: If all the certs in the base path change in one shot, multiple events will be triggered. If we consume all of these events one by one, multiple reload will be triggered. This `break` controls a list of events that will trigger only once reload.

src/main/java/org/tikv/common/util/ChannelFactory.java
iosmanthus: https://docs.oracle.com/javase/7/docs/api/java/nio/file/WatchService.html#take() should always return notnull seem this null check is redundant.

src/main/java/org/tikv/common/util/ChannelFactory.java
iosmanthus: I will wrap SslContextBuilder into a type with epoch.

iosmanthus:
@xuanyu66 PTAL

xuanyu66:
@iosmanthus Plz check the failed CI tests

src/main/java/org/tikv/common/util/ChannelFactory.java
```diff
@@ -26,80 +26,198 @@
...
+      try {
+        WatchService watchService = FileSystems.getDefault().newWatchService();
+        Path p = Paths.get(path);
+        p.register(watchService, StandardWatchEventKinds.ENTRY_MODIFY);
+
+        while (true) {
+          WatchKey key = watchService.take();
+          boolean changed = false;
+          OUTER:
+          for (WatchEvent<?> event : key.pollEvents()) {
```
xuanyu66, line 138: key.pollEvents() is non-blocking, which maybe waste CPU resource

src/main/java/org/tikv/common/util/ChannelFactory.java
```diff
@@ -212,72 +319,122 @@ public ChannelFactory(
...
   }
 
-  public ManagedChannel getChannel(String addressStr, HostMapping hostMapping) {
-    if (reloadSslContext()) {
-      logger.info("invalidate connection pool");
-      connPool.clear();
+  private void cleanExpireConn(Collection<ManagedChannel> pending) {
+    for (ManagedChannel channel : pending) {
+      logger.info("cleaning expire channels");
+      channel.shutdownNow();
```
xuanyu66, line 368: Why use shutdownNow()? Maybe some channel is being used right now

src/main/java/org/tikv/common/util/ChannelFactory.java
iosmanthus: this method will not keep running, once the events were taken out of the queue, this block will exist and wait at `watchService.take()`

src/main/java/org/tikv/common/util/ChannelFactory.java
iosmanthus: If the channel is used right now, no matter calling shutdown or `shutdownNow`, the user of the channel will receive an exception, a retry is expected.

iosmanthus:
> @iosmanthus Plz check the failed CI tests

Fixed, it's broken by the wrong initial state of `ChannelFactory`.

src/main/java/org/tikv/common/util/ChannelFactory.java
```diff
@@ -26,80 +26,198 @@
...
+    }
 
-    public SslContextBuilder reload() {
-      if (isModified()) {
-        logger.info("reload ssl context");
-        return createSslContextBuilder();
+    public boolean isModified() throws Exception {
+      boolean modified = !notify.isEmpty();
+      if (modified) {
+        notify.take();
```
zhangyangyu, line 173: Maybe better to use `clear`? If there are multiple `true`s in it, no need to trigger reload multiple times.

src/main/java/org/tikv/common/util/ChannelFactory.java
```diff
@@ -212,72 +319,122 @@ public ChannelFactory(
...
+      sslContextBuilder.set(new Epoch<>(certContext.createSslContextBuilder(), epoch++));
+
+      Collection<ManagedChannel> pending = new HashSet<>();
+      for (Map.Entry<Pair<Epoch<SslContextBuilder>, String>, ManagedChannel> pair :
+          connPool.entrySet()) {
+        Epoch<SslContextBuilder> b = pair.getKey().getLeft();
+        if (b != sslContextBuilder.get()) {
+          pending.add(pair.getValue());
+          connPool.remove(pair.getKey());
+          connPool.computeIfAbsent(
```
zhangyangyu, line 393: I don't get why we need to `computeIfAbsent` twice in reload.

zhangyangyu:
Given the complexity of current implementation, how do you folks think of using the simplest way to achieve it? A polling thread polls at fixed interval, like [org.apache.hadoop.security.ssl. ReloadingX509TrustManager](https://github.com/apache/hadoop/blob/d4fd675a95c16f6ecd8d8514cc3c0bef34bc9eff/hadoop-common-project/hadoop-common/src/main/java/org/apache/hadoop/security/ssl/ReloadingX509TrustManager.java). Make it optional let users open it and set interval. >_<!

iosmanthus:
> Given the complexity of current implementation and IMHO it's hard to reason about/understand/maintain(hard to make concurrency programming right), also break change the TLS configurations and `ChannelFactory` API :(, how do you folks think of using the simplest way to achieve it? A polling thread polls at fixed interval, like [org.apache.hadoop.security.ssl.ReloadingX509TrustManager](https://github.com/apache/hadoop/blob/d4fd675a95c16f6ecd8d8514cc3c0bef34bc9eff/hadoop-common-project/hadoop-common/src/main/java/org/apache/hadoop/security/ssl/ReloadingX509TrustManager.java). Make it optional, let users open it and set interval. >_<!

All the complexity come from the conn pool, even with a simple poll loop to refresh SSL context.

src/main/java/org/tikv/common/util/ChannelFactory.java
xuanyu66: got it

src/main/java/org/tikv/common/util/ChannelFactory.java
xuanyu66: Got it

xuanyu66 (dismissed):

iosmanthus:
@zhangyangyu PTAL

src/main/java/org/tikv/common/util/ChannelFactory.java
```diff
@@ -188,22 +199,39 @@ public ChannelFactory(
...
         new OpenSslContext(trustCertCollectionFilePath, keyCertChainFilePath, keyFilePath);
+
+    File trustCert = new File(trustCertCollectionFilePath);
+    File keyCert = new File(keyCertChainFilePath);
+    File key = new File(keyFilePath);
+
+    if (certReloadInterval > 0) {
+      onCertChange();
+      this.certWatcher =
+          new CertWatcher(1, ImmutableList.of(trustCert, keyCert, key), this::onCertChange);
```
shiyuhang0, line 222: why we set interval as 1 in openssl

src/main/java/org/tikv/common/util/ChannelFactory.java
```diff
@@ -188,22 +199,39 @@ public ChannelFactory(
...
+      this.certWatcher = null;
+    }
   }
 
   public ChannelFactory(
       int maxFrameSize,
       int keepaliveTime,
       int keepaliveTimeout,
       int idleTimeout,
+      long connRecycleTime,
```
shiyuhang0, line 233: I don't understand why a connRecycleTime delay is needed when recycling connections. Can you explain this configuration

src/main/java/org/tikv/common/util/ChannelFactory.java
```diff
@@ -40,66 +47,93 @@
...
-    public SslContextBuilder reload() {
-      if (isModified()) {
-        logger.info("reload ssl context");
-        return createSslContextBuilder();
+    private boolean needReload() {
+      boolean needReload = false;
+      // Check all the modification of the `targets`.
+      // If one of them changed, means to need reload.
+      for (int i = 0; i < targets.size(); i++) {
+        long lastModified = targets.get(i).lastModified();
```
zhangyangyu, line 101: `lastModified` might raise error due to permission. Maybe we need to catch the whole `needReload` body to give a meaningful error log?

src/main/java/org/tikv/common/util/ChannelFactory.java
```diff
@@ -320,8 +239,46 @@ public ChannelFactory(
...
+  }
+
+  private void onCertChange() {
+    try {
+      SslContextBuilder newBuilder = certContext.createSslContextBuilder();
+      lock.writeLock().lock();
+      sslContextBuilder.set(newBuilder);
+
+      List<ManagedChannel> pending = new ArrayList<>(connPool.values());
+      recycler.schedule(() -> cleanExpireConn(pending), connRecycleTime, TimeUnit.SECONDS);
```
xuanyu66: Since recycler is single-thread and prev task is still running even if cleaning was done.
So the interval of the current task is n*connRecycleTime(n is the number of tasks in the recycler).
https://blog.csdn.net/weixin_38106322/article/details/104494106

src/main/java/org/tikv/common/util/ChannelFactory.java
iosmanthus: It's a miss typing, change to `certReloadInterval`.

src/main/java/org/tikv/common/util/ChannelFactory.java
iosmanthus: Comments will be added, and the old channel will work for a while even if the certificate is reloaded but it should not be used forever, this recycle time is about to force those expired channels to close and use the latest context after a specific waiting duration.

shiyuhang0: get

src/main/java/org/tikv/common/util/ChannelFactory.java
```diff
@@ -212,72 +240,110 @@ public ChannelFactory(
...
+
+  private void cleanExpiredConn(List<ManagedChannel> pending) {
+    for (ManagedChannel channel : pending) {
+      logger.info("cleaning expire channels");
+      channel.shutdownNow();
+      while (!channel.isShutdown()) {
+        try {
+          channel.awaitTermination(5, TimeUnit.SECONDS);
+        } catch (Exception e) {
+          logger.info("recycle channels timeout:", e);
```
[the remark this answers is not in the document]
iosmanthus: Changed to `warn`.

src/main/java/org/tikv/common/region/StoreHealthyChecker.java
```diff
@@ -96,6 +96,7 @@ private boolean checkStoreTombstone(TiStore store) {
         return true;
       }
     } catch (Exception e) {
+      logger.info("fail to check tombstone stores {}", e.toString());
```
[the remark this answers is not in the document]
iosmanthus: Done.

src/main/java/org/tikv/common/util/ChannelFactory.java
iosmanthus: Done.

iosmanthus:
@zhangyangyu @xuanyu66 @sunxiaoguang @shiyuhang0 PTAL

xuanyu66 (approved):

shiyuhang0 (approved):

src/main/java/org/tikv/common/util/ChannelFactory.java
```diff
@@ -26,80 +26,198 @@
...
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.tikv.common.HostMapping;
 import org.tikv.common.pd.PDUtils;
 
 public class ChannelFactory implements AutoCloseable {
   private static final Logger logger = LoggerFactory.getLogger(ChannelFactory.class);
+  private static final String PUB_KEY_INFRA = "PKIX";
 
+  private int connRecycleTime;
```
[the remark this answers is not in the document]
iosmanthus: Done.

sunxiaoguang (dismissed):

sunxiaoguang (approved):

PR #617, opened 2022-06-17
[close #616] check store version while using API V2

### LINKED ISSUE #616: Check store version bound while creating V2 client
## Feature Request

The older version of the TiKV cluster (<= v6.1.0) allows the request from the newer V2 client to write which might break some assumptions of API V2 protocol like the format of the region (its bounds are encoding in memory-comparable format). But with version 6.1.0 TiKV, the check is added. We should add some store version checks while constructing the `TiSession`.

iosmanthus:

Issue Number: close #616

Check version bound in TiSession ctor.

### Code changes

### Check List for Tests

- Unit test
- Integration test

### Related changes

### DIFF
--- a/src/main/java/org/tikv/common/TiSession.java
+++ b/src/main/java/org/tikv/common/TiSession.java
@@ -175,6 +175,10 @@ public class TiSession implements AutoCloseable {
     }
 
     this.client = PDClient.createRaw(conf, keyCodec, channelFactory);
+    if (conf.getApiVersion().isV2() && !StoreVersion.minTiKVVersion(Version.API_V2, client)) {
+      throw new IllegalStateException("Store version should larger than " + Version.API_V2);
+    }
+
     this.enableGrpcForward = conf.getEnableGrpcForward();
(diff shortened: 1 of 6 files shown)


pingyu (dismissed):
LGTM~

src/main/java/org/tikv/common/TiSession.java
```diff
@@ -175,6 +175,10 @@ public TiSession(TiConfiguration conf) {
     }
 
     this.client = PDClient.createRaw(conf, keyCodec, channelFactory);
+    if (conf.getApiVersion().isV2() && !StoreVersion.minTiKVVersion(Version.API_V2, client)) {
+      throw new IllegalStateException("Store version should larger than " + Version.API_V2);
```
pingyu, line 179:
```suggestion
      throw new IllegalStateException("Store version should not be older than " + Version.API_V2);
```

pingyu (approved):

sunxiaoguang:
LGTM

ti-srebot (approved):

PR #618, opened 2022-06-20
[close #619] Shutdown recycler when closing ChannelFactory to avoid resource leak

### LINKED ISSUE #619: ScheduledExecutorService may not be shutdown in ChannelFactory
### 1. Describe the bug
```
public class ChannelFactory implements AutoCloseable {
    private final ScheduledExecutorService recycler = Executors.newSingleThreadScheduledExecutor();
     ....
     public void close() {
     ....
      if (certContext != null) {
        recycler.shutdown();                                      // recycler would not shutdown when TLS is disable
        if (certWatcher != null) {
          certWatcher.close();
        }
      }
    }
}
```

Daemonxiao:
Signed-off-by: Daemonxiao <735462752@qq.com>

Issue Number: close #619

Because the `recycler` in `ChannelFactory` does not set any task when TLS is turned off, the process will get stuck on exit.

When we don't use TLS, `recycler` should be `null`.

### Code changes

### Side effects

- NO side effects

### Related changes

### DIFF
--- a/src/main/java/org/tikv/common/util/ChannelFactory.java
+++ b/src/main/java/org/tikv/common/util/ChannelFactory.java
@@ -64,7 +64,7 @@ public class ChannelFactory implements AutoCloseable {
 
   private final AtomicReference<SslContextBuilder> sslContextBuilder = new AtomicReference<>();
 
-  private final ScheduledExecutorService recycler = Executors.newSingleThreadScheduledExecutor();
+  private final ScheduledExecutorService recycler;
 
   private final ReadWriteLock lock = new ReentrantReadWriteLock();
 
@@ -209,6 +209,7 @@ public class ChannelFactory implements AutoCloseable {
     this.idleTimeout = idleTimeout;
     this.certWatcher = null;
     this.certContext = null;
+    this.recycler = null;
     this.connRecycleTime = 0;
   }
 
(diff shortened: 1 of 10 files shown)


iosmanthus:
@Daemonxiao Could you create an issue with this pull request? The PR title checker will check if the PR is related to an issue.

src/main/java/org/tikv/common/util/ChannelFactory.java
```diff
@@ -361,7 +364,7 @@ public void close() {
     connPool.clear();
 
     if (certContext != null) {
-      recycler.shutdown();
+      recycler.shutdownNow();
```
iosmanthus, line 367: What about removing the `if (certContext != null)`:

```java
if (recycler != null ) {
    // shutdown recycler
}

if (certWatcher != null) {
    // shutdown certWatcher
}
```

Daemonxiao: @iosmanthus Thanks for your review. OK, I will change it.

iosmanthus:
The tests for `ChannelFactory` might leak executors because some `certWatcher` or `ChannelFactory` are not closed automatically with the try-with-resource block. Could you fix it within this pull request? I think these two issues are related. Rest LGTM

zhangyangyu (approved):

iosmanthus (approved):
LGTM

PR #624, opened 2022-06-27
[fix #623] region may be missed

### LINKED ISSUE #623: Table not found when TiDB has a lots of tables in a DB
### 1. Describe the bug

This bug is from TiSpark. [Issue2368](https://github.com/pingcap/tispark/issues/2368)
If we read a table after create a lots of table, it may throw table not find exception.

1. run a script that create 10000 tables with complex schema.
2. try to read a table when running the script, like  ` select count(*) from tbl5000`.
3. read the last table after finishing the script, like  `select count(*) from tbl10000`.

The second query throws a table not find exception.

A normal answer like:
+--------+
| count(1)|
+--------+
|       0      |
+--------+

- Client Java: we test it in TiSpark.
- TiKV: 5.4.0

TrafalgarRicardoLu:

Issue Number: close #623

Problem Description: If there is a region split between two queries, it may miss some regions.

I change the time of getting region from before scan to after scan.

When we build client, we will use region info in regionCache. After region split, it's outdated.
But when client try to scan the region, it will get the newest region info from pb. 

In the original code, we get region info before scan. So the region info is still outdated and it will use the old endKey to get next region. Using the old endKey may miss some regions that split during the two query.

I get the region info after scan to make sure the return value is newest.

### Check List for Tests

- Unit test
- Integration test

### DIFF
--- a/src/main/java/org/tikv/common/operation/iterator/ConcreteScanIterator.java
+++ b/src/main/java/org/tikv/common/operation/iterator/ConcreteScanIterator.java
@@ -77,6 +77,10 @@ public class ConcreteScanIterator extends ScanIterator {
       client.setTimeout(conf.getScanTimeout());
       BackOffer backOffer = ConcreteBackOffer.newScannerNextMaxBackOff();
       currentCache = client.scan(backOffer, startKey, version);
+      // If we get region before scan, we will use region from cache which
+      // may have wrong end key. This may miss some regions that split from old region.
+      // Client will get the newest region during scan. So we need to
+      // update region after scan.
       region = client.getRegion();
       return region;
     }
--- a/src/main/java/org/tikv/common/operation/iterator/ConcreteScanIterator.java
+++ b/src/main/java/org/tikv/common/operation/iterator/ConcreteScanIterator.java
@@ -75,9 +75,9 @@ public class ConcreteScanIterator extends ScanIterator {
     TiRegion region;
     try (RegionStoreClient client = builder.build(startKey)) {
       client.setTimeout(conf.getScanTimeout());
-      region = client.getRegion();
       BackOffer backOffer = ConcreteBackOffer.newScannerNextMaxBackOff();
       currentCache = client.scan(backOffer, startKey, version);
+      region = client.getRegion();
       return region;
     }
   }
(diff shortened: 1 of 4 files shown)


shiyuhang0:
@iosmanthus PTAL

src/main/java/org/tikv/common/operation/iterator/ConcreteScanIterator.java
```diff
@@ -75,9 +75,9 @@ TiRegion loadCurrentRegionToCache() throws GrpcException {
     TiRegion region;
     try (RegionStoreClient client = builder.build(startKey)) {
       client.setTimeout(conf.getScanTimeout());
-      region = client.getRegion();
       BackOffer backOffer = ConcreteBackOffer.newScannerNextMaxBackOff();
       currentCache = client.scan(backOffer, startKey, version);
+      region = client.getRegion();
```
iosmanthus, line 80: We might add some comments about this update since the `scan` has some implicit side effects. In addition, this issue might occur in https://github.com/tikv/client-java/blob/9ce52d34017e2caaba1a74335e454a87ebeed2af/src/main/java/org/tikv/common/operation/iterator/RawScanIterator.java#L49, the raw scan of client-java. We might add some mock tests for this interface, at least a **TODO** should be added.

TrafalgarRicardoLu: I have add comment and fix it in RawScanIterator. I think test it is not easy, so a TODO is added.

iosmanthus (dismissed):
LGTM

zhangyangyu (approved):

ti-srebot (approved):

sunxiaoguang:
Thanks for your contribution!

Could you please add test code to pass the code coverage check, thanks.

src/main/java/org/tikv/common/region/RegionStoreClient.java
```diff
@@ -1253,6 +1254,9 @@ public List<KvPair> rawScan(BackOffer backOffer, ByteString key, int limit, bool
               regionManager, this, resp -> resp.hasRegionError() ? resp.getRegionError() : null);
       RawScanResponse resp =
           callWithRetry(backOffer, TikvGrpc.getRawScanMethod(), factory, handler);
+      // RegionErrorHandler may refresh region cache due to outdated region info,
```
iosmanthus, line 1257: Seems this update is not necessary, since the region should be updated by the `RegionErrorHandler`. Same as transaction scan?

iosmanthus (approved):
LGTM

PR #638, opened 2022-07-28
[close #639] reduce lock granularity of RegionStoreClient and PDClient

### LINKED ISSUE #639: `PDClient` and `RegionStoreClient` are slow down massively in heavy workload.
### 1. Describe the bug

Every request will create a `RegionStoreClient` by asking the `RegionManager` about the region info before it is sent to TiKV servers.
https://github.com/tikv/client-java/blob/5a757d95e527822ac5bd52603290ec996d9a0f9a/src/main/java/org/tikv/common/region/RegionStoreClient.java#L1430-L1435

However, this method is declared synchronized, which might block the entire client every time it launches a request. It's safe to remove the `synchronized` keyword since the underlying code of `RegionManager` is already synced.

Another code path that suffers from the `synchronized` is:

https://github.com/tikv/client-java/blob/3128161b1f04a7088406b2b79f8a07d6144abab3/src/main/java/org/tikv/common/PDClient.java#L521

This code is executed when there is a request error or response error while interacting with PD servers. Under heavy retry, this code path might be slowed down by the lock. It is also safe to remove the synchronized keyword since the PD server is not required to update to the latest one, we might use `AtomicReference` to wrap and update the `pdClientWrapper`. Even if we get a stale leader in this code path, there is a thread that constantly updates the PD leader every 10 seconds.

1. Launch a workload with 32 threads.
2. Kill all the PD servers and TiKV servers
3. You will find two types of slow logs that are blocked in the code paths above.

### 3. What are your Java Client and TiKV versions? (Required)

- Client Java: [master](https://github.com/tikv/client-java/commit/3128161b1f04a7088406b2b79f8a07d6144abab3)
- TiKV: any version

iosmanthus:

Issue Number: close #639

Problem Description:

This pull request tries to reduce the lock granularity of `ReionStoreClient` and `PDClient`, we might block the requests while under a heavy workload.

1. Remove the synchronized keyword for RegionStoreBuilder.build, since the region cache's methods have already been synced.
2. Remove the synchronized keyword in PDClient's update leader logic, since the leader will be eventually updated to the latest one by the update leader scheduler.

### Check List for Tests

- Unit test
- Integration test

### Related changes

### DIFF
--- a/src/main/java/org/tikv/common/PDClient.java
+++ b/src/main/java/org/tikv/common/PDClient.java
@@ -523,8 +523,8 @@ public class PDClient extends AbstractGRPCClient<PDBlockingStub, PDFutureStub>
 
   public void tryUpdateLeaderOrForwardFollower() {
     if (updateLeaderNotify.compareAndSet(false, true)) {
-      BackOffer backOffer = ConcreteBackOffer.newCustomBackOff(BackOffer.PD_INFO_BACKOFF);
       try {
+        BackOffer backOffer = defaultBackOffer();
         updateLeaderService.submit(
             () -> {
               try {
--- a/src/main/java/org/tikv/common/PDClient.java
+++ b/src/main/java/org/tikv/common/PDClient.java
@@ -52,6 +52,7 @@ import java.util.concurrent.ConcurrentMap;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
+import java.util.concurrent.RejectedExecutionException;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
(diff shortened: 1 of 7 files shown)


src/main/java/org/tikv/common/PDClient.java
```diff
@@ -470,11 +471,12 @@ private GetMembersResponse getMembers(BackOffer backOffer, URI uri) {
   }
 
   // return whether the leader has changed to target address `leaderUrlStr`.
-  synchronized boolean trySwitchLeader(String leaderUrlStr) {
-    if (pdClientWrapper != null) {
-      if (leaderUrlStr.equals(pdClientWrapper.getLeaderInfo())) {
+  @VisibleForTesting
+  public boolean trySwitchLeader(String leaderUrlStr) {
```
xuanyu66, line 475: This method is not Thread-safe, since the value of pdClientWrapper can be changed during statements

src/main/java/org/tikv/common/PDClient.java
```diff
@@ -551,7 +554,8 @@ && createLeaderClientWrapper(leaderUrlStr)) {
 
       // If we have not used follower forward, try the first follower.
       boolean hasReachNextMember =
-          pdClientWrapper != null && pdClientWrapper.getStoreAddress().equals(leaderUrlStr);
+          pdClientWrapper.get() != null
```
xuanyu66, line 557: The same as above, loss of atomicity

src/main/java/org/tikv/common/PDClient.java
```diff
@@ -560,7 +564,8 @@ && createLeaderClientWrapper(leaderUrlStr)) {
         }
         String followerUrlStr = member.getClientUrlsList().get(0);
         followerUrlStr = uriToAddr(addrToUri(followerUrlStr));
-        if (pdClientWrapper != null && pdClientWrapper.getStoreAddress().equals(followerUrlStr)) {
+        if (pdClientWrapper.get() != null
```
xuanyu66, line 567: The same as above, loss of atomicity

xuanyu66 (changes requested):

sunxiaoguang (approved):

PR #645, opened 2022-08-09
[fix #643] fix retry with resolvedlocks info

### LINKED ISSUE #643: In client-java, is there a retry mechanism for batchGet Retry?
### 1. Describe the bug

In client-java, TiKVExeption exception "Lock not resolved, retry" Is it not retrying as this pullrequest does?  [#1908](https://github.com/pingcap/tispark/pull/1908/files)

Is there a BatchGet retry mechanism? (maybe the test part of this pr is clearer than mine is described)

yegetables:

Issue Number: close #643 

Problem Description: 

The lock information that is being cleaned up will be cached in the RegionStoreClient so that it can be used to unlock the next time it is retried. Failure to cache this information will cause the unlock to fail until the lock is cleaned up until it times out or the lock is cleaned up.

Like this pullrequest pingcap/tispark#1907, cache the lock information and try again to avoid creating a new client that does not hold the lock information to retry.

### Check List for Tests
- Unit test
- Integration test

### DIFF
--- a/src/main/java/org/tikv/common/region/RegionStoreClient.java
+++ b/src/main/java/org/tikv/common/region/RegionStoreClient.java
@@ -298,7 +298,16 @@ public class RegionStoreClient extends AbstractRegionStoreClient {
             forWrite);
     BatchGetResponse resp =
         callWithRetry(backOffer, TikvGrpc.getKvBatchGetMethod(), request, handler);
-    return handleBatchGetResponse(backOffer, resp, version);
+    try {
+      return handleBatchGetResponse(backOffer, resp, version);
+    } catch (TiKVException e) {
+      if ("locks not resolved, retry".equals(e.getMessage())) {
+        backOffer.doBackOff(BackOffFunction.BackOffFuncType.BoTxnLock, e);
+        return batchGet(backOffer, keys, version);
+      } else {
+        throw e;
+      }
+    }
   }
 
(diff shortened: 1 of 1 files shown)


src/main/java/org/tikv/common/region/RegionStoreClient.java
```diff
@@ -298,7 +298,16 @@ public List<KvPair> batchGet(BackOffer backOffer, List<ByteString> keys, long ve
             forWrite);
     BatchGetResponse resp =
         callWithRetry(backOffer, TikvGrpc.getKvBatchGetMethod(), request, handler);
-    return handleBatchGetResponse(backOffer, resp, version);
+    try {
+      return handleBatchGetResponse(backOffer, resp, version);
+    } catch (TiKVException e) {
+      if ("locks not resolved, retry".equals(e.getMessage())) {
```
iosmanthus, line 304: We should not retry depending on the error message. could we retry in `handleBatchGetResponse`?

ajian2002: handleBatchGetResponse may need to add a parameter `keys`

ajian2002:
Is this modification necessary to increase testing? @iosmanthus

shiyuhang0:
closed by duplicating with https://github.com/tikv/client-java/pull/685

PR #653, opened 2022-09-22
[close #654] Add RegionCacheInvalidCallBack

### LINKED ISSUE #654: Add a CallBack interface that allows users to customize the behavior when requesting region data fails
## Feature Request

Now the cache invalid envet is all handle by client-java. We want add a CallBack interface that allows users to customize the behavior when requesting region data fails.
In TiSpark, a Job execution may involve multiple machines, and the regionMananger on each machine may have its own invalid region. Through CallBack, we can collect the invalid regions from all machines and perform invalid processing on each machine to refresh them. This reduces the probability that a subsequent Job will encounter a region failure.

qidi1:
Signed-off-by: qidi1 <1083369179@qq.com>

Issue Number: close #654 

Problem Description: **TBD**

Add RegionCacheInvalidCallBack. Let the upper layers customize their own behavior when the region cache fails.

### Code changes

- No code

### Check List for Tests

- Unit test
### Side effects

- NO side effects

### Related changes

- NO related changes


iosmanthus:
Please open an issue to describe it.

src/main/java/org/tikv/common/operation/RegionErrorHandler.java
```diff
@@ -42,6 +44,7 @@
   private final Function<RespT, Errorpb.Error> getRegionError;
   private final RegionManager regionManager;
   private final RegionErrorReceiver recv;
+  private final Function<CacheInvalidateEvent, Void> cacheInvalidateCallBack;
```
xuanyu66, line 47: Using list is more resonable

xuanyu66 (changes requested):

src/main/java/org/tikv/common/operation/RegionErrorHandler.java
```diff
@@ -264,8 +269,23 @@ public TiRegion getRegion() {
...
-    regionManager.invalidateStore(ctxRegion.getLeader().getStoreId());
+    regionManager.invalidateStore(storeId);
+    notifyRegionStoreCacheInvalidate(
+        ctxRegion.getId(), storeId, CacheInvalidateEvent.CacheType.REGION_STORE);
+  }
+
+  private void notifyRegionStoreCacheInvalidate(
+      long regionId, long storeId, CacheInvalidateEvent.CacheType type) {
+    if (cacheInvalidateCallBack != null) {
+      cacheInvalidateCallBack.apply(new CacheInvalidateEvent(regionId, storeId, true, true, type));
```
xuanyu66, line 282: should catch error, callback must not break the main routine.
Also, it should be executed synchronizedly

src/main/java/org/tikv/common/operation/RegionErrorHandler.java
qidi1: Already add

src/main/java/org/tikv/common/operation/RegionErrorHandler.java
qidi1: Already add

src/main/java/org/tikv/common/operation/RegionErrorHandler.java
```diff
@@ -161,6 +163,7 @@ public boolean handleRegionError(BackOffer backOffer, Errorpb.Error error) {
       // key requested is not in current region
       // should not happen here.
       ByteString invalidKey = error.getKeyNotInRegion().getKey();
+      notifyRegionCacheInvalidate(recv.getRegion().getId(), CacheType.REGION_STORE);
```
shiyuhang0, line 166: Why we need to invalid cache here

src/main/java/org/tikv/common/operation/RegionErrorHandler.java
```diff
@@ -264,8 +269,34 @@ public TiRegion getRegion() {
...
+
+  private void notifyRegionStoreCacheInvalidate(
+      long regionId, long storeId, CacheInvalidateEvent.CacheType type) {
+    if (cacheInvalidateCallBackList != null) {
+      for (Function<CacheInvalidateEvent, Void> cacheInvalidateCallBack :
+          cacheInvalidateCallBackList) {
+        cacheInvalidateCallBack.apply(
+            new CacheInvalidateEvent(regionId, storeId, true, true, type));
+      }
+    }
```
shiyuhang0, line 287: Do not need to catch exception?

src/main/java/org/tikv/common/operation/RegionErrorHandler.java
```diff
@@ -142,8 +146,6 @@ public boolean handleRegionError(BackOffer backOffer, Errorpb.Error error) {
           BackOffFunction.BackOffFuncType.BoServerBusy,
           new StatusRuntimeException(
               Status.fromCode(Status.Code.UNAVAILABLE).withDescription(error.toString())));
-      backOffer.doBackOff(
-          BackOffFunction.BackOffFuncType.BoRegionMiss, new GrpcException(error.getMessage()));
```
Daemonxiao, line 146: Why delete the `BoRegionMiss` backoffer here？

src/main/java/org/tikv/common/region/RegionManager.java
```diff
@@ -101,6 +106,15 @@ public ReadOnlyPDClient getPDClient() {
     return this.pdClient;
   }
 
+  public synchronized List<Function<CacheInvalidateEvent, Void>> getCacheInvalidateCallbackList() {
```
xuanyu66, line 109: Use Collections.synchronizedList() rather than synchronized.

src/main/java/org/tikv/common/operation/RegionErrorHandler.java
```diff
@@ -264,8 +269,34 @@ public TiRegion getRegion() {
...
-  private void invalidateRegionStoreCache(TiRegion ctxRegion) {
+  private void invalidateRegionStoreCache(TiRegion ctxRegion, long storeId) {
     regionManager.invalidateRegion(ctxRegion);
-    regionManager.invalidateStore(ctxRegion.getLeader().getStoreId());
+    regionManager.invalidateStore(storeId);
+    notifyRegionStoreCacheInvalidate(
+        ctxRegion.getId(), storeId, CacheInvalidateEvent.CacheType.REGION_STORE);
+  }
+
+  private void notifyRegionStoreCacheInvalidate(
```
xuanyu66, line 279: It's better to figure out whether we can asynchronizedly run callbacks.
Since these callbacks may cost a lot of time

xuanyu66 (changes requested):

src/main/java/org/tikv/common/operation/RegionErrorHandler.java
qidi1: There are two backoffer here, and the error in not cost by `BoRegionMiss`, So I delete this.

src/main/java/org/tikv/common/operation/RegionErrorHandler.java
qidi1: Add a new thread to run

src/main/java/org/tikv/common/region/RegionManager.java
qidi1: use CopyOnWriteList here

src/main/java/org/tikv/common/operation/RegionErrorHandler.java
qidi1: Change to catch

src/main/java/org/tikv/common/operation/RegionErrorHandler.java
qidi1: I delete this

src/main/java/org/tikv/common/operation/RegionErrorHandler.java
```diff
@@ -264,8 +268,48 @@ public TiRegion getRegion() {
...
+    switch (type) {
+      case REGION:
+      case LEADER:
+        event = new CacheInvalidateEvent(ctxRegion.getId(), 0, true, false, type);
+        break;
+      case REGION_STORE:
+        event = new CacheInvalidateEvent(ctxRegion.getId(), storeId, true, true, type);
+        break;
+      case REQ_FAILED:
+        event = new CacheInvalidateEvent(0, 0, false, false, type);
```
xuanyu66, line 283: magic number

xuanyu66 (changes requested):

src/main/java/org/tikv/common/operation/RegionErrorHandler.java
qidi1: remove magic number here

src/main/java/org/tikv/common/operation/RegionErrorHandler.java
```diff
@@ -264,8 +274,52 @@ public TiRegion getRegion() {
...
+    }
+    if (cacheInvalidateCallBackList != null) {
+      for (Function<CacheInvalidateEvent, Void> cacheInvalidateCallBack :
+          cacheInvalidateCallBackList) {
+        callBackThreadPool.submit(
+            () -> {
+              try {
+                cacheInvalidateCallBack.apply(event);
+              } catch (Exception e) {
+                logger.warn(String.format("CacheInvalidCallBack failed %s", e));
```
xuanyu66, line 304: logger.error
there is no need to throw exception

shiyuhang0 (dismissed):

PR #737, opened 2023-04-10
[close #736] Select reachable store

### LINKED ISSUE #736: Missing fallback mechanism for replica selection
### 1. Describe the bug

When there are multiple replica candidates, and the first one has already failed to respond, the caller does not try the next one on the list.

1. Use leader and follower replica selector
2. Use follower read
3. Kill the store that hosts the target follower before sending the request

As stated in the bug description.

Try the next replica candidate

- Client Java: 6.5
- TiKV: 6.5

Ideally, it is expected to see that a request trys to connect to the first replica candidate on the list, retry for several times or a certain period of time on the same target when it is unavailable (considering transient error and rate limiting are quite normal on the TiKV side), then fallback to the next one on the list if the first candidate is still unavailable.

shiyuhang0:

Issue Number: close #736

Judge store status before use. Use the next replica when the current replica is unreachable.

### Test

I have tested in TiSpark follower read with the first store killed, its success to request the next store.

### DIFF
--- a/src/main/java/org/tikv/common/region/RegionManager.java
+++ b/src/main/java/org/tikv/common/region/RegionManager.java
@@ -228,13 +228,16 @@ public class RegionManager {
 
     TiStore store = null;
     if (storeType == TiStoreType.TiKV) {
-      region.SetReplicaIdx(0);
-      Peer peer = region.getCurrentReplica();
-      store = getStoreById(peer.getStoreId(), backOffer);
-      while (!store.isReachable()) {
-        logger.info("Store is unreachable, try to get the next replica");
-        peer = region.getNextReplica();
+      // check from the first replica in case it recovers
+      for (int i = 0; i < region.getReplicaList().size(); i++) {
+        Peer peer = region.getReplicaList().get(i);
         store = getStoreById(peer.getStoreId(), backOffer);
+        if (store.isReachable()) {
+          // update replica index
+          region.setReplicaIdx(i);
+          break;
+        }
+        logger.info("Store is unreachable, try to get the replica,index: " + i);
       }
     } else {
       List<TiStore> tiflashStores = new ArrayList<>();
--- a/src/main/java/org/tikv/common/region/RegionManager.java
+++ b/src/main/java/org/tikv/common/region/RegionManager.java
@@ -232,6 +232,7 @@ public class RegionManager {
       Peer peer = region.getCurrentReplica();
       store = getStoreById(peer.getStoreId(), backOffer);
       while (!store.isReachable()) {
+        logger.info("Store is unreachable, try to get the next replica");
         peer = region.getNextReplica();
         store = getStoreById(peer.getStoreId(), backOffer);
       }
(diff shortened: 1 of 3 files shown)


zhangyangyu (approved):

src/main/java/org/tikv/common/region/RegionManager.java
```diff
@@ -228,12 +228,25 @@ public Pair<TiRegion, TiStore> getRegionStorePairByKey(
 
     TiStore store = null;
     if (storeType == TiStoreType.TiKV) {
-      Peer peer = region.getCurrentReplica();
-      store = getStoreById(peer.getStoreId(), backOffer);
+      // check from the first replica in case it recovers
+      List<Peer> replicaList = region.getReplicaList();
+      for (int i = 0; i < replicaList.size(); i++) {
+        Peer peer = replicaList.get(i);
+        store = getStoreById(peer.getStoreId(), backOffer);
```
pingyu, line 235: Is it possible to make replicas not balance ? As we always try the replica 0 at first.

How about start from `getCurrentReplica()`, then if the current replica is not available, get next one by `getNextReplica` and retry.

shiyuhang0: There are two problems if we getNextReplica.
1. code will trap into endless loop when there is no available store.
2. assume we set `follower,leader`, the replica list will begin with the followers then the leaders.  If we getNextReplica, we may miss the recover follower and request the leader

pingyu: > 1. code will trap into endless loop when there is no available store.

It is not difficult to be solved by loop no more than `replicaList.size()` times.

> 2. assume we set `follower,leader`, the replica list will begin with the followers then the leaders.  If we getNextReplica, we may miss the recover follower and request the leader

Why we should not "miss the recover follower and request the leader" ? If we don't want to read on leader, we can set the replica selector as "followers only". Otherwise, if "leader and follower" is used, it should be OK to read on leader.

> Besides, the original code always select the first one. also has the issue of balance

I also feel strange that why `getNextReplica` is unused, and except which `replicaIdx` is changed nowhere. Any idea @iosmanthus ?

zhangyangyu: The customer wants a "fallback" strategy. They want to read followers first to isolate dump traffic. But if all the followers are down, fallback to leaders as a cover.

So if "follower and leader" is specified, it means use followers first, and then leaders. The order matters.

pingyu: Is stale read enabled ? If not, when all followers are down, the leader is also unavailable (neither readable nor writable as well, see https://www.pingcap.com/blog/lease-read/)

zhangyangyu: Ahh, I think we take some terms wrong, it's not follower, it's learner, who will not attend voting. And yes, this is for stale read. The normal TiKV nodes will still take leader and follower and handle online traffic. And there will be extra learner nodes dedicated for offline traffic. When learners are down, goes for the normal nodes.

Issue #776, opened 2024-01-18
fix prewrite & commit log info to debug


jackjeyis:
### 1. Describe the bug

TwoPhaseCommit prewrite 、commit  log use info level 

### 3. What did you see instead (Required)

      LOG.info(
          "start commit secondary key, row={}, size={}KB, regionId={}",
          batchKeys.getKeys().size(),
          batchKeys.getSizeInKB(),
          batchKeys.getRegion().getId());

### 4. What did you expect to see? (Required)

    if (LOG.isDebugEnabled()) {
      LOG.debug(
          "start commit secondary key, row={}, size={}KB, regionId={}",
          batchKeys.getKeys().size(),
          batchKeys.getSizeInKB(),
          batchKeys.getRegion().getId());
    }

### 5. What are your Java Client and TiKV versions? (Required)

- Client Java: 3.3.5
- TiKV: 7.5.0

PR #775, opened 2024-01-18
[close #774] Add TwoPhaseCommitter new Constructor contains executorService(#774)

jackjeyis:
### What problem does this PR solve?

Issue Number: close #issue_number

Problem Description: **TBD**

### Code changes

- Has exported function/method change
- Has exported variable/fields change
- Has methods of interface change
- Has persistent data change
- No code

### Check List for Tests

This PR has been tested by at least one of the following methods:
- Unit test
- Integration test
- Manual test (add detailed scripts or steps below)
- No code

### Side effects

- Possible performance regression, WHY: **TBD**
- Increased code complexity, WHY: **TBD**
- Breaking backward compatibility, WHY: **TBD**
- NO side effects

### Related changes

- Need to cherry-pick to the release branch
- Need to update the documentation
- Need to be included in the release note
- NO related changes

PR #777, opened 2024-01-18
fix prewrite & commit log info to debug(#776)

jackjeyis:

Issue Number: close #issue_number

Problem Description: **TBD**

### Code changes

- Has methods of interface change
- Has persistent data change
- No code

### Check List for Tests

- Unit test
- Integration test
- No code

### Side effects

- Possible performance regression, WHY: **TBD**
- Increased code complexity, WHY: **TBD**
- Breaking backward compatibility, WHY: **TBD**
- NO side effects

### Related changes

- NO related changes

### DIFF
--- a/src/main/java/org/tikv/txn/TwoPhaseCommitter.java
+++ b/src/main/java/org/tikv/txn/TwoPhaseCommitter.java
@@ -227,8 +227,9 @@ public class TwoPhaseCommitter implements AutoCloseable {
         this.doCommitPrimaryKeyWithRetry(backOffer, key, commitTs);
       }
     }
-
-    LOG.info("commit primary key {} successfully", KeyUtils.formatBytes(key));
+    if (commitResult.isSuccess() && LOG.isDebugEnabled()) {
+      LOG.debug("commit primary key {} successfully", KeyUtils.formatBytes(key));
+    }
   }
 
   /**
@@ -397,11 +398,13 @@ public class TwoPhaseCommitter implements AutoCloseable {
       BatchKeys batchKeys,
       Map<ByteString, Kvrpcpb.Mutation> mutations)
       throws TiBatchWriteException {
-    LOG.debug(
-        "start prewrite secondary key, row={}, size={}KB, regionId={}",
-        batchKeys.getKeys().size(),
-        batchKeys.getSizeInKB(),
-        batchKeys.getRegion().getId());
+    if (LOG.isDebugEnabled()) {
+      LOG.debug(
+          "start prewrite secondary key, row={}, size={}KB, regionId={}",
+          batchKeys.getKeys().size(),
+          batchKeys.getSizeInKB(),
+          batchKeys.getRegion().getId());
+    }
 
     List<ByteString> keyList = batchKeys.getKeys();
     int batchSize = keyList.size();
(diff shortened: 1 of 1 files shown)


src/main/java/org/tikv/txn/TxnKVClient.java
```diff
@@ -181,7 +180,6 @@ public ClientRPCResult commit(
   // TODO: check this logic to see are we satisfied?
   private boolean retryableException(Exception e) {
     return e instanceof TiClientInternalException
-        || e instanceof KeyException
```
shiyuhang0, line 184: Delete this will be a break change. Some KeyException also need the retry, why do you want to exclude this?
If you still want to exclude some KeyException, please judge the type. here is an example
```
 if (e instanceof KeyException) {
      Kvrpcpb.KeyError ke = ((KeyException) e).getKeyError();
      if (ke == null) return true;
      if (!ke.getAbort().isEmpty()
          || ke.hasConflict()
          || ke.hasAlreadyExist()
          || ke.hasDeadlock()
          || ke.hasCommitTsExpired()
          || ke.hasTxnNotFound()) return false;
      return true;
    }
```

src/test/java/org/tikv/txn/TwoPhaseCommitterTest.java
```diff
@@ -76,4 +99,170 @@ public void autoClosableTest() throws Exception {
             executorService)) {}
     Assert.assertTrue(executorService.isShutdown());
   }
+
+  @Test
+  public void prewriteWriteConflictFastFailTest() throws Exception {
```
shiyuhang0, line 104: It seems this test has nothing to do with this pr, right?

shiyuhang0: This test is based on the new constructor method added in this PR https://github.com/tikv/client-java/pull/775. 
Do you mind move it to that PR, or the test will fail

jackjeyis: this test for verify write conflict will aways retry ,but need fast fail

src/main/java/org/tikv/txn/TxnKVClient.java
jackjeyis: ok，I review it again

src/main/java/org/tikv/common/exception/KeyException.java
```diff
@@ -31,7 +31,7 @@ public KeyException(String errMsg) {
   }
 
   public KeyException(Kvrpcpb.KeyError keyErr) {
-    super("Key exception occurred");
+    super("Key exception occurred " + keyErr.toString());
```
shiyuhang0, line 34: Why do you need to print error msg here? 
Have you tried `getKeyErr()`?  it should return the error and you should be able to get the error msg from `getKeyErr().toString()`

src/test/java/org/tikv/txn/TwoPhaseCommitterTest.java
shiyuhang0: could you explain why this test was added? Or what scenes/feature this test for

PR #793, opened 2026-04-29
Fix #756: Update the `scan` method to include the end key

### LINKED ISSUE #756: KVClient scan error
### 1. KVClient scan error

It occurred exception when using the following method of org.tikv.txn.KVClient.
public List<Kvrpcpb.KvPair> scan(ByteString startKey, ByteString endKey, long version) throws GrpcException;

1) Insert test data into TiKV using TwoPhaseCommitter successfully
Test data contains several KV pairs, and all the keys start with "h[tikv-client-java]txn1`"

2) Use KVClient.scan to get all the data of which the keys start with "h[tikv-client-java]txn1`"
KVClient.scan(ByteString.copyFromUtf8("h[tikv-client-java]txn1`"),
                        ByteString.copyFromUtf8("h[tikv-client-java]txn1a"), transactionStartTimestamp);

Scan failed with exceptions like
org.tikv.common.exception.TiClientInternalException: Error scanning data from region.
org.tikv.common.exception.KeyException: tikv abort txn Error(InvalidKeyRangeMode { cmd: scan, storage_api_version: V2, range: (Some("78000000685B7469FF6B762D636C69656EFF742D6A6176615D74FF786E316000000000FB"), None) })

image

image

All the data of which the keys start with "h[tikv-client-java]txn1`"

- Client Java:  3.3.4
- TiKV:  v6.1.1

wot4n:
Issue Number: #756 

Fix Txn API v2 scan: bound-aware RPC and decoded keys
 * `RegionStoreClient.scan` now supports an endKey overload. API v2 builds the scan via `codec.encodeRange(startKey, endKey)` and sets StartKey + EndKey on the KV scan request (matching keyspace-encoded bounds). API v1 keeps the prior behavior (start only, no end on this path).
* Scan responses are returned through `codec.decodeKvPairs(...)` so iterators see logical user keys, aligned with batchGet and correct prefix/end semantics under v2 encoding.
* ConcreteScanIterator / ScanIterator carry the raw range end (rangeEndKey) into scan(...) together with keyOnly.
* KVMockServer: honor request.getEndKey() when simulating bounded scans.

Previously, scan only sent a wired start without a wired end. That could (a) fetch more data than necessary and (b) return pairs whose Key bytes still carried the prefix, which did not line up with batchGet and other APIs that expose logical keys after decoding.

The fix (a) sends a proper EndKey on the KV scan for v2 so TiKV can bound the read in storage coordinates, and (b) decodes pairs so upper layers always see user-space keys, consistent with the rest of the client.

`ConcreteScanIterator` still advances startKey region-to-region using decoded keys from `currentCache` (now decoded). Passing `rangeEndKey` into scan keeps each batch limited on the RPC for v2 instead of relying only on client-side filtering after pulling an unbounded stream from TiKV where the RPC had no end bound.

### Code changes
- Has exported function/method change. 
 
I added a new public `scan` method to `src/main/java/org/tikv/common/region/RegionStoreClient.java`. This new method is similar to the existing one but the main difference is that the new one adds the `ByteString endKey` method argument while the old one is using an empty values using `ByteString.EMPTY`.

### Check List for Tests
- Integration test
- I published a local version of the lib to my local maven repository to test it and it is working as expected.

### Side effects
- NO side effects. I added a new method while maintaining compatibility with the old method.

### Related changes
- NO related changes

## Summary by CodeRabbit

* **New Features**
  * Added support for bounded range scans with explicit end-key specification.

* **Bug Fixes**
  * Scan operations now correctly enforce end-key boundaries when iterating and returning results.

* **Chores**
  * Project artifact version updated to 3.4.0 for release.


wot4n:
Kindly review @iosmanthus

